PR #10, opened 2017-01-20
vpView needs vidtk to be usable

RustyBlue:
@mattdawkins PTAL


mattdawkins:
This will temporarily break the vivia option in master, but it will be fixed by a shortcoming patch

PR #11, opened 2017-01-22
Add Burn-Out, Update VXL Module, Update Vivia Module

mattdawkins:


CMake/add_project_vibrant.cmake
```diff
@@ -14,11 +14,12 @@ set( VIBRANT_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x" )
 ExternalProject_Add(vibrant
   DEPENDS fletch
   PREFIX ${VIAME_BUILD_PREFIX}
-  SOURCE_DIR ${VIAME_PACKAGES_DIR}/vibrant
+  SOURCE_DIR ${VIAME_PACKAGES_DIR}/burn-out
```
linus-sherrill: Is this the right directory for vibant?

mattdawkins: Nope, but burn-out is a fork of vibrant just a lot more modern. I'll clean this up in a little bit, and replace add_project_vibrant with add_project_burnout or the like, merging it now just so that vpView works in master (currently track loading functionality is broken).

PR #15, opened 2017-02-13
Remove vibrant subpackage

mattdawkins:
Remove vibrant in favor of burn-out

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,9 +7,6 @@
 [submodule "packages/scallop-tk"]
 	path = packages/scallop-tk
 	url = https://github.com/mattdawkins/scallop-tk.git
-[submodule "packages/vibrant"]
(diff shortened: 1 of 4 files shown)


.gitmodules
```diff
@@ -7,9 +7,6 @@
 [submodule "packages/scallop-tk"]
 	path = packages/scallop-tk
 	url = https://github.com/mattdawkins/scallop-tk.git
-[submodule "packages/vibrant"]
-	path = packages/vibrant
-	url = https://github.com/Kitware/vibrant.git
 [submodule "packages/vivia"]
```
linus-sherrill: I would expect to see the burnout submodule being added

mattdawkins: It was already added in prior commit, ie, right now they are both checked in and one needs to be removed to prevent code duplication.

PR #17, opened 2017-03-22
Adding windows support for caffe

hughed2:


hughed2:
The way we link libraries for caffe into Scallop-TK needs to be fixed up, but this will work to push out the door in the short term and shouldn't affect non-Scallop-TK and non-Windows builds. An issue is that it relies on ntdll.lib, which I can supply but I don't believe comes standard on Windows and is a pain in the butt to get (it's part of WDK, which requires a larger install and can screw other settings).

CMakeLists.txt
```diff
@@ -120,10 +120,10 @@ if( VIAME_ENABLE_CUDNN AND NOT VIAME_ENABLE_CUDA )
...
 endif()
 
-if( WIN32 AND VIAME_ENABLE_CAFFE )
-  message( FATAL_ERROR "Internal Caffe build on windows not currently supported, \
-    please build Caffe externally using the Caffe windows branch, if desired." )
-endif()
+#if( WIN32 AND VIAME_ENABLE_CAFFE )
+#  message( FATAL_ERROR "Internal Caffe build on windows not currently supported, \
+#    please build Caffe externally using the Caffe windows branch, if desired." )
+#endif()
```
linus-sherrill: If this is just dead code, could it be removed?

hughed2: Yes it could.

hughed2:
Pending some passing tests, I've got a better fix coming in for the linking situation. I'll leave the script as is tonight and leave the PR open in case Matt wants to use this tomorrow morning, but linking gods willing there will be a new version up fairly soon.

mattdawkins:
I've merged this into master through other means

PR #20, opened 2017-05-25
Add -D flags that disable Eigen alignment stuff for 32 bit builds.

billhoffman:


billhoffman:
@mattdawkins please merge to fix 32 bit builds.

PR #25, opened 2017-09-13
Refactor camtrawl matlab into python and C++ code, add measurement capabilities

Erotemic:
Preliminary work on integration of camtrawl pipeline with VIAME, KWIVER and Sprokit. 

The algo implementations are in `plugins/camtrawl/python/camtrawl_algos.py` 
A standalone demo (that does not use sprokit) is in `plugins/camtrawl/python/camtrawl_demo.py` 

The initial sprokit integration is in `plugins/camtrawl/python/camtrawl_process.py`


CMake/add_project_kwiver.cmake
```diff
@@ -73,14 +73,18 @@ ExternalProject_Add(kwiver
     -DKWIVER_ENABLE_DARKNET:BOOL=${VIAME_ENABLE_YOLO}
     -DKWIVER_ENABLE_C_BINDINGS:BOOL=${VIAME_ENABLE_PYTHON}
     -DKWIVER_ENABLE_PYTHON:BOOL=${VIAME_ENABLE_PYTHON}
+    -DKWIVER_SYMLINK_PYTHON:BOOL=${VIAME_SYMLINK_PYTHON}
     -DKWIVER_ENABLE_CUDA:BOOL=${VIAME_ENABLE_CUDA}
     -DKWIVER_ENABLE_DOCS:BOOL=${VIAME_ENABLE_DOCS}
     -DKWIVER_INSTALL_DOCS:BOOL=${VIAME_ENABLE_DOCS}
     -DKWIVER_ENABLE_TRACK_ORACLE:BOOL=${VIAME_ENABLE_KWANT}
+    -D KWIVER_ENABLE_TESTS:BOOL=On
```
mattdawkins: KWIVER enable tests should not be enforced to always on, we should add a high level viame flag which also controls this

plugins/camtrawl/matlab/rodrigues.m
```diff
@@ -0,0 +1,267 @@
...
+%%
+%%		Copyright (c) March 1993 -- Pietro Perona
+%%		California Institute of Technology
+%%
+
+%% ALL CHECKED BY JEAN-YVES BOUGUET, October 1995.
+%% FOR ALL JACOBIAN MATRICES !!! LOOK AT THE TEST AT THE END !!
+
+%% BUG when norm(om)=pi fixed -- April 6th, 1997;
+%% Jean-Yves Bouguet
```
mattdawkins: Watch the line ending types (even though this isn't your code)

CMake/add_project_kwiver.cmake
linus-sherrill: The space between D and K may cause a problem

Erotemic: How so? I've always used a space between -D and the variable when writing scripts that invoke cmake, but after looking at all the internal code I see the overwhelming pattern is to not have the space. I'm curious under what conditions would the space cause problems?

Purg: If nothing else, we may as well be self-consistent.

mattdawkins:
camtrawl2 has been merged into master, this can be abandoned, right?

Erotemic:
correct

PR #26, opened 2017-10-02
Dev/add lanl

mattdawkins:


mattdawkins:
@lprasad02 I just cleaned up a few things (e.g. cmakelists)

PR #28, opened 2017-10-20
option to disable forcebuild for easier development

Erotemic:
I was trying to build with all dependencies and I ran into an issue with smqtk, and it was a nightmare to try and fix the issue having to wait for VTK and all the other fletch projects to build before I could check if my smqtk fix worked. 

I recalled a conversation with @mattdawkins about disabling forced builds, and I found a way to do it. It is pretty hacky, but its simple enough where I think its fine as a stop-gap measure / developer-trick. 

All the `ExternalProject_Add_Step(<target> forcebuild...`  commands were basically the same, so I abstracted them all into a separate function. In this function I allowed a global variable called `VIAME_FORCEBUILD_<target>` to control whether or not they actually were force-built. 

I populate these variables as options in the main CMakeLists depending on which projects are enabled. Thus the user just flips one of these variables to false, and it wont automatically remove its build stamp anymore. 

Because it seems kind of dangerous / hacky to do this, these options are marked as advanced. 
I took care to ensure that any package without one of these variables defined would still be force-built, and I ensured that the viame project is never force built (nor is it even given the option).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@ project( VIAME )
 
 include( CMakeDependentOption )
 
+
 ###
 # Add paths used throughout CMakeLists files
 ##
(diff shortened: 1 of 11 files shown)


mattdawkins:
Superseded by https://github.com/VIAME/VIAME/pull/292.

The current build already has the shared force-build helper and `VIAME_BUILD_FORCE_REBUILD` switch, which defaults to OFF. The remaining gap was Fletch's unconditional `BUILD_ALWAYS` setting.

The replacement adds `VIAME_BUILD_FLETCH_ALWAYS` (default ON) so developers can opt out of repeated Fletch builds without disabling the project. It preserves first-build behavior and the existing global force override, and documents the option. These behaviors were checked with both Makefiles and Ninja using a small dependency fixture.

Closing this older PR in favor of the focused replacement against current main. Neither PR is being merged as part of this update.

PR #29, opened 2017-10-21
Fixed issue with installing smqtk when a virtualenv is enabled

Erotemic:
The pip `--user` option doesn't work if you use a python virtualenv. 

While I did find a workaround for old versions of pip, they were extremely cumbersome and required me to preinstall `psycopg2` before I tried to install `smqtk`. 

Thinking about it, I figured any person using a virtualenv is probably going to have a more recent version of pip. Thus, I wrote a function that checks the version of pip and if its modern (`>=9.0.0`). If its not it uses the `--user` + environment variable method to install the package to a specific directory. Any virtualenvs users will still fail here, but hopefully they will just `pip install pip -U` and be on their way.  

It the pip version is `>=9.0.0`, then the smqtk install command uses the much more concise and standard `--prefix` synatax. This version also uses the simplified extra_requires syntax as well without needing the `file://` stuff.    

While I was here I also put in code to determine and use the correct version of site-packages / dist-packages when setting up the PYTHONPATH.  This uses some copy/pasted logic from kwiver (although I did use an improved version of my pycmd function.)

### DIFF
--- a/cmake/add_project_smqtk.cmake
+++ b/cmake/add_project_smqtk.cmake
@@ -9,29 +9,84 @@
 
 set( VIAME_PROJECT_LIST ${VIAME_PROJECT_LIST} smqtk )
 
+
+###
+# Private helper function to execute `python -c "<cmd>"`
+#
+# Runs a python command and populates an outvar with the result of stdout.
+# Be careful of indentation if `cmd` is multiline.
+#
+function(_pycmd outvar cmd)
+  execute_process(
+    COMMAND "${PYTHON_EXECUTABLE}" -c "${cmd}"
+    RESULT_VARIABLE _exitcode
+    OUTPUT_VARIABLE _output)
+  if(NOT ${_exitcode} EQUAL 0)
+    message(ERROR "Failed when running python code: \"\"\"
+${cmd}\"\"\"")
+    message(FATAL_ERROR "Python command failed with error code: ${_exitcode}")
+  endif()
+  # Remove supurflous newlines (artifacts of print)
+  string(STRIP "${_output}" _output)
+  set(${outvar} "${_output}" PARENT_SCOPE)
+endfunction()
+
+# Copy logic for getting site packages from kwiver
+_pycmd(python_site_packages "\
+from distutils import sysconfig
+print(sysconfig.get_python_lib(prefix=''))
+")
+message(STATUS "python_site_packages = ${python_site_packages}")
+get_filename_component(python_sitename ${python_site_packages} NAME)
+
+_pycmd(python_pip_version "import pip; print(pip.__version__)")
+
 if( WIN32 )
   message( FATAL_ERROR "SMQTK not yet supported on WIN32" )
 else()
-  if( VIAME_SYMLINK_PYTHON )
-    set( SMQTK_PIP_CMD
-      pip install --user -e .[postgres] )
-  else()
-    # This is only required for no symlink install without a -e with older
-    # versions of pip, for never versions the above command works with no -e
-    set( SMQTK_PIP_CMD
-      pip install --user file://${VIAME_PACKAGES_DIR}/smqtk\#egg=smqtk[postgres] )
-  endif()
+
   set( PYTHON_BASEPATH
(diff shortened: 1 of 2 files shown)


cmake/add_project_smqtk.cmake
```diff
@@ -12,26 +12,60 @@ set( VIAME_PROJECT_LIST ${VIAME_PROJECT_LIST} smqtk )
...
-      pip install --user -e .[postgres] )
-  else()
-    # This is only required for no symlink install without a -e with older
-    # versions of pip, for never versions the above command works with no -e
-    set( SMQTK_PIP_CMD
-      pip install --user file://${VIAME_PACKAGES_DIR}/smqtk\#egg=smqtk[postgres] )
-  endif()
+
+  # Logic for getting site packages is copied from kwiver
+  # (but it uses a new pycmd function that handles indentation -- ooohh. aahh.)
```
Purg: Less oohh and aahh?  Not sure how others feel about flavorful comments.

cmake/add_project_smqtk.cmake
```diff
@@ -12,26 +12,60 @@ set( VIAME_PROJECT_LIST ${VIAME_PROJECT_LIST} smqtk )
...
+    endif()
+    set( SMQTK_PYTHON_INSTALL
+      ${CMAKE_COMMAND} -E env PYTHONPATH=${CUSTOM_PYTHONPATH}
+                          env PATH=${CUSTOM_PATH}
+                          env PYTHONUSERBASE=${VIAME_BUILD_INSTALL_PREFIX}
+        ${PYTHON_EXECUTABLE} -m ${SMQTK_PIP_CMD} )
+  else()
+    # If we are using virtualenvs then a more up to date pip is necessary
+    if( VIAME_SYMLINK_PYTHON )
+      set( SMQTK_PIP_CMD pip install "--prefix=${VIAME_BUILD_INSTALL_PREFIX}" -e .[postgres])
```
Purg: Instead of setting SMQTK_PIP_CMD independently in both if and else, since this ``if`` is just adding the ``-e``, we can set that string in a variable and unpack it after the statement:
```cmake
...
set( pip_args "" )
if( VIAME_SYMLINK_PYTHON )
    set( pip_args "-e" )
endif()
set( SMQTK_PIP_CMD pip install "--prefix=${VIAME_BUILD_INSTALL_PREFIX}" ${pip_args} .[postgres] )
...
```

cmake/add_project_smqtk.cmake
Erotemic: consider it a minor easter egg for the git archives (unless we rebase and squash).

cmake/add_project_smqtk.cmake
Erotemic: I would agree if the `-e` was a standalone argument that changed the behavior of pip. However, if you see the pip docs:
```
  -e, --editable <path/url>   Install a project in editable mode (i.e. setuptools "develop mode") from a local project path or a VCS
```
the `-e` is followed by a path, meaning that if futher pip args needed to be specified (e.g. `--no-deps`), then the code would break if that were appended to the end of the args. 

The alternative is to do something like this: 
```cmake
set( pip_args "" )
if( VIAME_SYMLINK_PYTHON )
    set( pip_args "-e .[postgres]" )
else()
    set( pip_args ".[postgres]" )
endif()
set( SMQTK_PIP_CMD pip install "--prefix=${VIAME_BUILD_INSTALL_PREFIX}" ${pip_args} )
```

or this

```cmake
set( pip_path ".[postgres]" )
if( VIAME_SYMLINK_PYTHON )
    set( pip_path "-e ${pip_path}" )
endif()
set( SMQTK_PIP_CMD pip install "--prefix=${VIAME_BUILD_INSTALL_PREFIX}" ${pip_path} )
```
However, I feel like both of these are less clear than the way it exists now.

Purg: Ahh. I missed the semantic that ``-e`` required a path after it.  What's here is fine then.

mattdawkins:
I am reverting this out of next and re-submitting the PR

PR #31, opened 2017-10-23
Fix build issue when system GDAL is installed. 

Erotemic:
Requires Kitware/burn-out#2

Fixes a built issue if system GDAL is installed


mattdawkins:
I'm abandoning this branch since it has already been made on master (The burnout_disable_gdal flag was also changed to ENABLE_GDAL for consisitency

PR #39, opened 2018-02-22
Fixes for python virtualenv

Erotemic:
Updates to fix Python 3 builds as I find them

So far I have:
 
* Fixed some cmake flags
* Fixed SMQTK


mattdawkins:
Seeing build errors with this on windows, also the branch shouldn't contain large swaths of commented code if going into master/next

CMakeLists.txt
```diff
@@ -123,25 +123,28 @@ if( VIAME_ENABLE_PYTHON )
 
   option( VIAME_SYMLINK_PYTHON "Symlink python files instead of copying." OFF )
   mark_as_advanced( VIAME_SYMLINK_PYTHON )
+  # set flag so we can use kwiver python utils
+  set(KWIVER_SYMLINK_PYTHON VIAME_SYMLINK_PYTHON)
```
mattdawkins: This variable is set but not used?

CMakeLists.txt
```diff
@@ -352,8 +355,8 @@ else()
 
   # Setup python directories for plugin processes in same style as KWIVER
   if( VIAME_ENABLE_PYTHON )
-    set( kwiver_python_subdir "python${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" )
-    set( kwiver_python_output_path "${VIAME_BUILD_PREFIX}/lib/${kwiver_python_subdir}" )
+    pycmd(python_site_packages "from distutils import sysconfig; print(sysconfig.get_python_lib(prefix=''))")
+    set( kwiver_python_output_path "${VIAME_BUILD_PREFIX}/${python_site_packages}" )
```
mattdawkins: This makes the path on windows odd lib/Lib/python...

mattdawkins (changes requested):

CMakeLists.txt
mattdawkins: Further clarrification, it's not used at the super-build level, but must be set down in the VIAME subproject

Erotemic: KWIVER_SYMLINK_PYTHON should be set if you call kwiver_add_python_module, so it is used in the VIAME subproject. Did I put it in the wrong place?

CMakeLists.txt
Erotemic: `sysconfig.get_python_lib(prefix='')` should return `Lib\\site-packages` on windows. Is that not happening? What happens when you message the `python_site_packages` var?

CMakeLists.txt
mattdawkins: Yeah it needed to be in the VIAME subproject not at the super-build level (inside the large else statement lower down)

mattdawkins:
I changed this branch's target to be master now, but there are conflicts and unresolved issues (breaking outside of a virtual env).

CMakeLists.txt
mattdawkins: It puts the files in [install]/lib/Lib/... shouldn't there only be a single lib?

mattdawkins:
I cherry-picked the python3 stuff from this branch, but did not merge it due to breaking things using native python outside of a few virtualenvs, also now needs a rebase.

mattdawkins:
Closing as superseded. As noted in the earlier maintainer comment, the useful Python 3 changes were already cherry-picked, while the remaining virtualenv changes broke native Python configurations and had Windows build issues. Current main has since replaced the Python build/setup infrastructure, and the SMQTK integration targeted by this patch has been removed. There is no remaining change here that warrants merging or rebasing as-is; any current virtualenv issue should be addressed with a focused fix against main.

PR #40, opened 2018-02-22
Fix our smqtk build for windows

hughed2:
Using the default env paths breaks Windows. We could also sanitize the path, but this is quicker and easier for our deadline. We also need to get rid of those extra "env" flags, proper usage is "cmake -E env var1=value1 var2=value2 var3=value3".

### DIFF
--- a/cmake/add_project_smqtk.cmake
+++ b/cmake/add_project_smqtk.cmake
@@ -9,9 +9,6 @@
 
 set( VIAME_PROJECT_LIST ${VIAME_PROJECT_LIST} smqtk )
 
-if( WIN32 )
-  message( FATAL_ERROR "SMQTK not yet supported on WIN32" )
-else()
   if( VIAME_SYMLINK_PYTHON )
     set( SMQTK_PIP_CMD
       pip install --user -e .[postgres] )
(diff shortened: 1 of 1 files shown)


Erotemic:
this will have conflicts with #39 which includes fixes for smqtk when you have a virtualenv

cmake/add_project_smqtk.cmake
```diff
@@ -24,15 +21,14 @@ else()
...
-    ${VIAME_BUILD_INSTALL_PREFIX}/bin:$ENV{PATH} )
+    ${VIAME_BUILD_INSTALL_PREFIX}/bin )
   set( SMQTK_PYTHON_INSTALL
     ${CMAKE_COMMAND} -E env PYTHONPATH=${CUSTOM_PYTHONPATH}
-                        env PATH=${CUSTOM_PATH}
-                        env PYTHONUSERBASE=${VIAME_BUILD_INSTALL_PREFIX}
+                        PATH=${CUSTOM_PATH}
+                        PYTHONUSERBASE=${VIAME_BUILD_INSTALL_PREFIX}
       ${PYTHON_EXECUTABLE} -m ${SMQTK_PIP_CMD} )
-endif()
```
mattdawkins: Why do you still have the first 2 spaces if the if statement was removed?

mattdawkins: In the old version before his latest commit he had an extra indentation for the entire if block but was fixed and merged

PR #49, opened 2018-04-20
Add clustering annotation gui

hughed2:
Couple notes:

This isn't a huge PR. I don't know how you want to organize things, but if you want to change the directory structure it's a pretty trivial change.

This doesn't include any of the test videos or anything that they sent us. If you want to try them out, untar fishtracker.tar and move the contents of /docker/data to wherever you set FT_DATA_ROOT to (defaulting to $CMAKE_BIN_DIR/data).

I took out all the extraneous python scripts. They're in cd48087 if you ever want to take a look at them, they might be necessary if you want to use any of their tracking code instead of our own.


mattdawkins:
Looks like you don't have a CMakeList.txt which installs the plugin into the build/install tree?

mattdawkins:
Ah, I see you have a cmake install command at the top level. We should change it and put it inside the plugin but can do that after merge.

hughed2:
I'm planning on reorganizing this a little anyway. I've noticed that depending on some other settings, it either installs to and lets you set FT_DATA_ROOT relative to the top level, or in build/src/viame-build. The inconsistency's not good, I want to guarantee it going to the appropriate bin and python directories instead.

PR #77, opened 2019-03-15
Tensorflow plugins

neelsj:
Added tensor flow detector plugins with color normalization internal to plugin for loading 16 bit IR.


mattdawkins:
I'll probably make a couple minor changes tommorow quick unless you object but: changing the config params to snake case so they match the others in the configs, just having one root level tf library since it's pretty small, moving the pipeline files to configs/pipelines/detector_seal_tf_*, run scripts to examples.

neelsj:
No objections 

-Neel

—-
From my phone

PR #80, opened 2019-04-11
add cuda stubs dir for darknet

Cookt2:
Fixes this error I was getting when building in a docker container:

[27%] Performing configure step for 'darknet'
loading initial cache file /root/viame/build/build/tmp/darknet-cache-Release.cmake
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
CUDA_CUDA_LIBRARY (ADVANCED)
    linked by target "darknet_lib" in directory /root/viame/packages/darknet/src

-- Configuring incomplete, errors occurred!
See also "/root/viame/build/build/src/darknet-build/CMakeFiles/CMakeOutput.log".
See also "/root/viame/build/build/src/darknet-build/CMakeFiles/CMakeError.log".
CMakeFiles/darknet.dir/build.make:108: recipe for target 'build/src/darknet-stamp/darknet-configure' failed
make[2]: *** [build/src/darknet-stamp/darknet-configure] Error 1
CMakeFiles/Makefile2:109: recipe for target 'CMakeFiles/darknet.dir/all' failed
make[1]: *** [CMakeFiles/darknet.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2


cmake/add_project_darknet.cmake
```diff
@@ -17,6 +17,9 @@ endif()
 
 if( VIAME_ENABLE_CUDA )
   FormatPassdowns( "CUDA" VIAME_CUDA_FLAGS )
+  set( VIAME_DARKNET_CUDA_FLAGS 
+    -DCUDA_TOOLKIT_ROOT_DIR:PATH=${CUDA_TOOLKIT_ROOT_DIR}
```
mattdawkins: This CUDA path should already be included in the above macro, were you seeing issues with it being passed down?

cmake/add_project_darknet.cmake
```diff
@@ -17,6 +17,9 @@ endif()
...
+    -DCMAKE_LIBRARY_PATH:PATH=${CUDA_TOOLKIT_ROOT_DIR}/lib64/stubs )
```
mattdawkins: Should this be:

-DCMAKE_LIBRARY_PATH:PATH=${CUDA_TOOLKIT_ROOT_DIR}/lib64/stubs:${CMAKE_LIBRARY_PATH}

Legit question, I don't know the answer.

mattdawkins: Also as opposed to making a second variable, appending them to VIAME_CUDA_FLAGs or just changing VIAME_CUDA_FLAGS to VIAME_DARKNET_CUDA_FLAGS probably makes sense to prevent 2 variables from doing the same thing

Cookt2: > Should this be:
> 
> -DCMAKE_LIBRARY_PATH:PATH=${CUDA_TOOLKIT_ROOT_DIR}/lib64/stubs:${CMAKE_LIBRARY_PATH}
> 
> Legit question, I don't know the answer.

I don't think so, or at least I haven't seen needing to do that in any cmake code.

Cookt2: > Also as opposed to making a second variable, appending them to VIAME_CUDA_FLAGs or just changing VIAME_CUDA_FLAGS to VIAME_DARKNET_CUDA_FLAGS probably makes sense to prevent 2 variables from doing the same thing

My initial thought was I didn't want to append to VIAME_CUDA_FLAGS incase it was called again later on for other projects. (VIAME_CUDA_FLAGS sounds very general)

cmake/add_project_darknet.cmake
Cookt2: I observed that it wasn't getting passed down at all. Fletch had the same issue with building darknet, and this is the same exact patch it needed.

mattdawkins: It should be in the above macro, which searches the CMAKE cache for anything starting with CUDA_* and then passes them down, but can leave in for now and revist later

cmake/add_project_darknet.cmake
mattdawkins: It is used in other projects, but it gets over-written here and in everyone so it's only used internally to the file (though we could also call both DARKNET_CUDA flags, I'll make a quick commit after merge)

PR #86, opened 2019-07-01
Use TinyXML1 from Fletch, not just TinyXML

neal-siekierski:
As of commit e09a284 in Fletch (from May 16, 2019), merged into its
viame/master branch in commit 97ce409 (from June 24, 2019), the old
TinyXML defines are gone and replaced with TinyXML{1,2}.  TinyXML1
corresponds to the original, so we use that.


neal-siekierski:
I was running into issues with the Kwiver build because TinyXML wasn't getting built. This resolves that issue, though I don't know if it's the best way.

linus-sherrill:
Looks like a reasonable approach to me.

PR #92, opened 2019-08-07
Add dockerfile 

xkortex:
Adds dockerfile which uses some stages for limited caching. Still takes a very long time to build. 
Currently uses text files to allow for multiple configurations, however this will probably not pan out and will be replaced by dockerfiles for specific configurations.


xkortex:
I added a dockerfile for the Seal-TK loadout we were using before. It builds. The main dockerfile is still WIP.

.env
```diff
@@ -0,0 +1,4 @@
+GITSHA=latest
+VIAME_VARIANT=nodeps
+BASE_IMAGE=nvidia/cuda:9.1-cudnn7-devel-ubuntu16.04
+OPENCV_VERSION=4.1.0
```
mattdawkins: What is this file used for and is it necessary? (also the default cuda version is now 10.1, and OpenCV is still like 3.6 by default in what's used)

mattdawkins: (Main question is can I move it down to the docker directory, don't want to add many more at the root project level, same thing for docker compose)

xkortex: .env is used to fill in variables in the `docker-compose.yml` file. Since posting this MR, I have figured out a few more docker patterns, so I could rearrange things to avoid being at the root project level. I'll update the PR when I get some time.

PR #94, opened 2019-11-11
Fixed bug in python plugin template

SubodhDahal:
The plugin template for detector written in python doesn't work because it tries to access methods that are not present and also uses `print` statement which isn't present in Python 3. This PR is aimed to fix it. Please have a look.


mattdawkins:
Thanks, coincidentally I was just thinking about how out of date these templates were last week. I'll try to merge this and fix them later today.

PR #95, opened 2019-11-15
Add netharn detector

Erotemic:
Adds a module to wrap netharn trained detectors.


cmake/add_project_python_deps.cmake
```diff
@@ -12,11 +12,32 @@
...
+
+  list( APPEND VIAME_PYTHON_DEPS kwplot )
+  list( APPEND VIAME_PYTHON_DEP_CMDS "kwplot" )
+
+  list( APPEND VIAME_PYTHON_DEPS ndsampler )
+  list( APPEND VIAME_PYTHON_DEP_CMDS "ndsampler" )
+
+  list( APPEND VIAME_PYTHON_DEPS netharn )
+  list( APPEND VIAME_PYTHON_DEP_CMDS "netharn" )
+endif()
```
mattdawkins: Are specifying all of these necessary, or is it only necessary to include netharn and then all of the others are requirements in  netharn's  setup.py/requirements?

cmake/add_project_pytorch.cmake
```diff
@@ -37,9 +37,9 @@ if( VIAME_ENABLE_PYTORCH-PYSOT )
   set( PYTORCH_LIBS_TO_BUILD ${PYTORCH_LIBS_TO_BUILD} pysot )
 endif()
 
-if( VIAME_ENABLE_PYTORCH-NETHARN )
-  set( PYTORCH_LIBS_TO_BUILD ${PYTORCH_LIBS_TO_BUILD} netharn )
-endif()
+#if( VIAME_ENABLE_PYTORCH-NETHARN )
+#  set( PYTORCH_LIBS_TO_BUILD ${PYTORCH_LIBS_TO_BUILD} netharn )
+#endif()
```
mattdawkins: Would recommend removing dead code (and also removing the subpackage in packages/netharn if its now installed via above pip call)

plugins/pytorch/CMakeLists.txt
```diff
@@ -18,4 +18,14 @@ if( VIAME_ENABLE_PYTORCH-MMDET )
...
+  kwiver_add_python_module(
+    ${CMAKE_CURRENT_SOURCE_DIR}/netharn_detector.py
+    viame/arrows/pytorch
+    netharn_detector )
+
+  kwiver_add_python_module(
+    ${CMAKE_CURRENT_SOURCE_DIR}/_netharn_detect_predict.py
+    viame/arrows/pytorch
+    _netharn_detect_predict )
+
```
mattdawkins: Please put in VIAME_ENABLE_PYTORCH-NETHARN conditional

cmake/add_project_python_deps.cmake
Erotemic: I could remove kwarray and kwimage, but the other packages are not direct requirements of netharn (they are optional). However, they are needed for this PR. I think I'd prefer to leave kwimage in just due to the issues it will cause on non-linux systems in the near-term future. So I could remove kwarray, but maybe given that the others are necessary it would be better not to and make requirements more explicit?

In general you shouldn't trust that pip will install requirements correctly. It has some mechanism to find a reasonable ordering, but its heuristic based. It is not a general solution. Conda does have a general solution, but that involves solving an integer program, which can be a bit slow.

cmake/add_project_pytorch.cmake
Erotemic: Fixed.

plugins/pytorch/CMakeLists.txt
Erotemic: Fixed.

PR #97, opened 2019-12-11
Interactive container for persistent builds

xkortex:
Configured dockerfile and compose file for persistent viame builds. 
Uses named volumes to mount the repository dir and build dir, such that code updates and the build tree persist between up/down.

`/src` is currently mounted RO but not actively used in the build script in any way since fletch tries to download to the source tree.


xkortex:
Wrong target, ignore

PR #99, opened 2020-01-22
Teach read_detected_object_set_viame_csv to return file names

predicative:
The intent of the `file_name` output of the `detected_object_output` process in upstream Kwiver appears to be to return the file name associated with each batch of detected objects. However, `read_detected_object_set_viame_csv` simply passes the input file name it receives through, even when empty, which means that one cannot get the file names out without already knowing what they are.

This PR makes `read_detected_object_set_viame_csv` output file names when the input is empty (and therefore when it's iterating using frame numbers). It assumes that all detections with the same frame number name the same file; if not, it's the name used with the first occurrence of that frame number that is output.

The use case I have for this is preserving the file name information when using Kwiver/Sprokit to convert detection file formats.

### DIFF
--- a/plugins/core/read_detected_object_set_viame_csv.cxx
+++ b/plugins/core/read_detected_object_set_viame_csv.cxx
@@ -44,6 +44,8 @@
 
 #include <map>
 #include <sstream>
+#include <tuple>
+#include <utility>
 #include <cstdlib>
 
 namespace viame {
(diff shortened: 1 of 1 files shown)


mattdawkins:
Closing as already implemented. The current VIAME CSV reader records filenames in m_name_by_id and returns them when image_name is initially empty. Review validation of CSV-to-COCO conversion without supplying image files preserved the input filenames first.png and second.png. The behavior requested by this PR is already supported on main.

PR #101, opened 2020-02-13
Add converters for VIAME CSV <-> COCO

predicative:
This PR (currently) simply adds two pipeline files to convert VIAME CSV to and from the COCO format to the `detection_file_conversions` example.

Some possible points of attention:
- The other pipelines in this folder use an image list as input; these don't. (I'm not sure of the advantage of the former approach.)
- Related to the previous point, the CSV -> COCO converter requires #99 to work properly.
- I have a separate Python script, not included, to extract an image list from a COCO file, which is useful for creating an image list to accompany the CSV resulting from conversion. Do we also want this? If so, where? I have another script that runs both the conversion and the image list extraction, and again the question is where to put it.
- Is there a summary of the COCO format I could reference for this example's README? The format used does depart slightly from the one at http://cocodataset.org/#format-data, namely in the addition of a `"score"` attribute currently, but perhaps supporting more of the Kitware extensions later.

### DIFF
--- a/examples/detection_file_conversions/pipelines/coco_to_viame_csv.pipe
+++ b/examples/detection_file_conversions/pipelines/coco_to_viame_csv.pipe
@@ -1,0 +1,13 @@
+process read :: detected_object_input
+  file_name = input.coco.json
+  reader:type = DetectedObjectSetInputCoco
+
+process write :: detected_object_output
+  file_name = out.csv
+  writer:type = viame_csv
+
+connect from read.detected_object_set
+     	to write.detected_object_set
+
+connect from read.image_file_name
+     	to write.image_file_name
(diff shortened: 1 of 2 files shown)


mattdawkins:
The image list input is to ensure that the frame_ids in the csv (if used) match the position in the input list in case a custom ordering of input images is desired

I think you want to refer to this as a KWIVER .json format in the readme, but still say that it still supports native COCO (and link to it), as we're likely to add in tracking support if using it and brand it as that.

mattdawkins:
w.r.t. the whole binding setup it would be nice if the coco readers could support this (what the other ones do is if a file string (path) is passed into read_set, then it produces detections for that file)

mattdawkins:
deprecated

PR #108, opened 2020-08-17
Update mosaic tool

neal-siekierski:
This PR has all the changes I've made to the `create_mosaic.py` tool. Summarizing the most important changes:
- Added support for drawing any range of images with a consistent reference frame, not just the first block.
- Added `--optimize-fit` flag that attempts to minimize distortion of the input images in the output.
- Added `--reverse` flag that draws earlier frames on top of later frames.
- Added `--zoom` option that scales the output image (most useful with values less than unity in order to reduce memory consumption).
- Added Python function `main_multi` that can draw multiple co-registered image sequences (not available at the command line).

And a couple bug fixes:
- Fixed rounding-related bug that caused crashes for some inputs (and removed associated workaround).
- The progress bar now will go to 100%, though only after the rendered mosaic is written to disk.

### DIFF
--- a/examples/image_registration/create_mosaic.py
+++ b/examples/image_registration/create_mosaic.py
@@ -3,7 +3,7 @@ import glob
 import itertools as itt
 
 import numpy as np
-from scipy import ndimage
+import scipy.optimize
 from skimage import (
     io as skio,
     transform as sktr,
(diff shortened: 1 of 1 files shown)


mattdawkins:
I'd recommend moving this python script to the high-level tools directory, that's where I've putting putting standalone useful python scripts and C++ execs

neal-siekierski:
> I'd recommend moving this python script to the high-level tools directory, that's where I've putting putting standalone useful python scripts and C++ execs

Done. For what it's worth, it doesn't look like its dependencies are being handled quite right; it requires `numpy`, `scipy`, `scikit-image`, and `tqdm`. To judge by my reading of [`cmake/add_project_python_deps.cmake`](https://github.com/VIAME/VIAME/blob/952451d51c34fbbdd6ff59e6ca552614fc4f783c/cmake/add_project_python_deps.cmake), `numpy` is always installed, and `scipy` and `tqdm` are installed when `VIAME_ENABLE_OPENCV` is set, which is the condition under which the script is installed, but the `scikit-image` requirement is only installed when `VIAME_ENABLE_PYTORCH` is set. But that was already an issue, and it could be fixed at some other point.

PR #113, opened 2020-10-02
Added polygon to VIAME-csv reader

russelldj:
I have not gotten a chance to test this one its own, but it should be the same as what was done previously in the context of reading and writing.

### DIFF
--- a/plugins/core/CMakeLists.txt
+++ b/plugins/core/CMakeLists.txt
@@ -1,5 +1,20 @@
 # Build / Install Plugin containing core algorithm implementations
 
+# Find VXL
+find_package( VXL REQUIRED )
+include( ${VXL_CMAKE_DIR}/UseVXL.cmake )
+include_directories( SYSTEM ${VXL_CORE_INCLUDE_DIR} )
+include_directories( SYSTEM ${VXL_VCL_INCLUDE_DIR} )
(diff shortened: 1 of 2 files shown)


plugins/core/CMakeLists.txt
```diff
@@ -74,6 +93,7 @@ find_package( Boost ${KWIVER_BOOST_VERSION} REQUIRED
 include_directories( SYSTEM ${Boost_INCLUDE_DIRS} )
 link_directories( ${Boost_LIBRARY_DIRS} )
 
+
```
mattdawkins, line 96: why the extra line

plugins/core/read_detected_object_set_viame_csv.cxx
```diff
@@ -323,6 +337,56 @@ ::read_all()
       dob = std::make_shared< kwiver::vital::detected_object>( bbox, conf );
     }
 
+    if ( m_read_poly && col[i][0] == '(' ){
```
mattdawkins, line 340: generally you'd want to match styles of the code being modified, bracket start/end in same fashion, honoring total line spacing and not adding extra, etc; here and below

mattdawkins:
I added a bit more on c65a776ffccbb18064b5b8f488eb948cf1721705 if wanted to spot check that, added support for multiple polygons and extra checking for dependencies

PR #116, opened 2020-11-23
Make current directory evaluation in setup_viame.sh work outside of bash

russelldj:
I built VIAME but when I tried to `source setup_viame.sh` using `zsh` instead of `bash`, it failed because the `$BASH_SOURCE` variable is empty. Running `bash setup_viame.sh` runs without error, but does not export the required variables into the calling shell. `./setup_viame.sh` fails at the substitution stage for the shell prompt. 

I replaced `$BASH_SOURCE` with the more generic `$0`, which has worked for me on `bash` and `zsh`.  I've been using `zsh` on linux, which I don't think is super common, but it is now the default shell for MacOS starting with Catalina so it seems useful to support.


russelldj:
Both `$BASH_SOURCE[0]` and `$0` fail when using the `fish` shell. It would be nice is there was a POSIX compliant solution but I feel like anyone using a non-standard shell like that can deal with the issue.

mattdawkins:
Replaced by #293, built against current main. The Zsh sourcing issue remains relevant, but the original setup files have been replaced and using $0 can locate the caller instead of the sourced setup script in Bash. The replacement uses explicit Bash/Zsh source-file detection, shares it across the current development and release templates, and handles quoted installation paths without GNU readlink -f. Both Bash and Zsh CTest tests pass, covering 48 sourcing/configuration cases. Closing this PR in favor of that tested replacement.

PR #122, opened 2021-01-11
Convert detections to tracks and write

russelldj:
This updates the `detector_local.pipe` to write out track files in addition to detections. This is because the detection writer doesn't currently write timestamps.

### DIFF
--- a/configs/pipelines/detector_local.pipe
+++ b/configs/pipelines/detector_local.pipe
@@ -26,10 +26,34 @@ process detector_writer
   :file_name                                   computed_detections.csv
   :writer:type                                 viame_csv
 
+process track_initializer
+  :: initialize_object_tracks
+  :track_initializer:type                      threshold
+
+  block track_initializer:threshold:filter
+   :type                                      class_probablity_filter
+   :class_probablity_filter:threshold         0.0
+   :class_probablity_filter:keep_all_classes  true
+  endblock
+
+process track_writer
+  :: write_object_track
+  :file_name                                   computed_tracks.csv
+  :writer:type                                 viame_csv
+  :writer:viame_csv:active_writing             false
+
 connect from downsampler.output_1
         to   detector_input.image
 connect from downsampler.output_2
         to   detector_writer.image_file_name
+connect from downsampler.timestamp
+        to   track_initializer.timestamp
+
+connect from classifier1.detected_object_set
+        to   track_initializer.detected_object_set
+
+connect from track_initializer.object_track_set
+        to   track_writer.object_track_set
 
 connect from detector_output.detected_object_set
         to   detector_writer.detected_object_set
--- a/configs/pipelines/detector_local.pipe
+++ b/configs/pipelines/detector_local.pipe
@@ -30,7 +30,7 @@ process track_initializer
    :class_probablity_filter:keep_all_classes  true
   endblock
 
-process track_writer
+process detector_writer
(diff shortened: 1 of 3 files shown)


configs/pipelines/detector_local.pipe
```diff
@@ -26,10 +26,34 @@ process detector_writer
   :file_name                                   computed_detections.csv
   :writer:type                                 viame_csv
 
+process track_initializer
+  :: initialize_object_tracks
+  :track_initializer:type                      threshold
+
+  block track_initializer:threshold:filter
+   :type                                      class_probablity_filter
+   :class_probablity_filter:threshold         0.0
```
mattdawkins, line 35: Minor: Inconstant spacing with rest of parameters (looks off by 2)

mattdawkins (commented):
Also needs same application to full_frame_classifier_local.pipe

mattdawkins:
We also don't want both a computed tracks file and computed detections, that's excess when they will contain the same things, just want one or the other

mattdawkins:
We also need this branch to still function when processing an image list and have the 2nd column be the image name instead of the timestamp; to that end process_videos.py needs to be updated to set the correct writer param when this is a video vs an image list, on whether the 2nd column should be one vs the other

russelldj:
> We also need this branch to still function when processing an image list and have the 2nd column be the image name instead of the timestamp; to that end process_videos.py needs to be updated to set the correct writer param when this is a video vs an image list, on whether the 2nd column should be one vs the other
> 
> Note: this might already be in place and just need verification but I'm not certain

It currently looks like [this line](https://github.com/VIAME/VIAME/blob/341473b571420e941f807b86d93c9990f44f8ba3/tools/process_video.py#L771) selects whether to run with `write_time_as_uid`. However, this is targeted at `track_writer` which I had to update to `detection_writer`. I'd almost argue we should change both options back to `track_writer` for more clarity, but it works as is..

russelldj:
I think there may be an issue in the `write_object_track_set_viame_csv` for image lists which uses the image list file as the second column, rather than the actual image filename. However, it's possible I just didn't configure something correctly.

russelldj:
> I think there may be an issue in the `write_object_track_set_viame_csv` for image lists which uses the image list file as the second column, rather than the actual image filename. However, it's possible I just didn't configure something correctly.

While I didn't do anything to intentionally fix this, the issue is now resolved with the latest commit.

russelldj:
This issue with `active_writing`, which only produced one line of classifications, was related to [this line in the writer](https://github.com/VIAME/VIAME/blob/49cc5bb5c3edb1714d5dbf009adede12b180753c/plugins/core/write_object_track_set_viame_csv.cxx#L381l). I had failed to connect a timestamp port to the writer, and this led to the timestep for the current frame always being `0`. However, this information  isn't used directly in the output so it failed silently. I wonder if perhaps this should at least be a warning if a timestamp is wrong? Or can this process introspect to see if there's not timestamp port connected and fail for the `active_writing` case?

Upon further inspection, it looks like failing to connect the timestamp also created the issue above. [This line](https://github.com/VIAME/VIAME/blob/49cc5bb5c3edb1714d5dbf009adede12b180753c/plugins/core/write_object_track_set_viame_csv.cxx#L342) requires an active timestamp to populate `m_frame_uids`. When this variable is empty, the second column defaults to the stream identifier.

PR #125, opened 2021-01-29
Add polygon writing support for viame_csv detection output

neal-siekierski:
This teaches `write_detected_object_set_viame_csv` a configuration option `mask_to_poly_tol` that, when ~~not `-999`~~ nonnegative, makes it write masks as polygons. I've attempted to make it scale meaningfully to differently sized objects by scaling the value by the shorter of the object's width and height. It's then used as the `epsilon` parameter to [`cv::approxPolyDp`](https://docs.opencv.org/master/d3/dc0/group__imgproc__shape.html#ga0012a5fdaea70b8a9970165d98722b4c).

**Edit:** This now also adds a configuration option `mask_to_poly_points` that when nonnegative takes a similar approach to simplification (that is, also based on the [Ramer&ndash;Douglas&ndash;Peucker algorithm](https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm)), but specifies a maximum number of points instead. It's an error to specify both configuration options.

~~This also fixes an issue with the class's `set_configuration` method where the new configuration was being merged too late to be used to configure the object.~~ **Edit:** This issue was fixed separately in 010f5d1ec43a8750bb1c9be11328f11c4c0c5519. This PR now only does the cleanup it had done along with the fix.

### DIFF
--- a/plugins/core/write_detected_object_set_viame_csv.cxx
+++ b/plugins/core/write_detected_object_set_viame_csv.cxx
@@ -34,6 +34,13 @@
 
 #include <vital/util/tokenize.h>
 
+#ifdef VIAME_ENABLE_OPENCV
+  #include <arrows/ocv/image_container.h>
+
+  #include <opencv2/core/core.hpp>
+  #include <opencv2/imgproc/imgproc.hpp>
+#endif
+
 #include <memory>
 #include <vector>
 #include <fstream>
@@ -61,6 +68,7 @@ public:
     , m_stream_identifier( "" )
     , m_model_identifier( "" )
     , m_version_identifier( "" )
+    , m_mask_to_poly_tol( -1 )
   {}
 
   ~priv() {}
@@ -72,6 +80,7 @@ public:
   std::string m_stream_identifier;
   std::string m_model_identifier;
   std::string m_version_identifier;
+  double m_mask_to_poly_tol;
 };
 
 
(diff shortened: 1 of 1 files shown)


mattdawkins:
Minor comments: why the static helper function outside of the namespace as opposed to inside? Minor comments about bracket style inconsistencies after for loops on some line in places and not in others, will fix in post production

neal-siekierski:
> Minor comments: why the static helper function outside of the namespace as opposed to inside?

It means one less `#ifdef` block, but otherwise I don't think I had any particular reason. It's static &ndash; local to the file &ndash; so it doesn't pollute the global namespace. Feel free to move it.

> Minor comments about bracket style inconsistencies after for loops on some line in places and not in others, will fix in post production

You're right; it looks like I missed one [here](https://github.com/VIAME/VIAME/pull/125/files#diff-c0d43ce5c1d3a5d65d032f028252b1e9c59c45e35bd064d5d9e97eb00280c673R376). The opening brackets for the lambdas being on the same line follows KWIVER's [style guide's](https://github.com/Kitware/kwiver/blob/6e188289d0af35a15cf83e6601434f07a89a4c98/doc/code-style.rst) "\[indent.braces.lambda\]" guideline as I understand it, for what that's worth.

PR #126, opened 2021-01-29
Add pipes to add segmentations with OpenCV's watershed

neal-siekierski:
This depends on #125 and also needs Kitware/kwiver#1194 to be integrated. I've based this PR on #125's branch for clarity, but it shouldn't actually be merged into it.

This adds a file `viame_csv_add_segmentations.pipe` that uses the `ocv_watershed` implementation of `refine_detections` from Kitware/kwiver#1194. I've placed under the `detection_file_conversions` example, but there's probably a better place to put it.

It also adds `detector_add_segmentations.pipe` under `configs/pipelines`. This is substantially the same as the other one but is compatible with DIVE and is set to read in polygons from its input file.

`viame_csv_add_segmentations.pipe` could probably be removed. Perhaps we also want to add a GrabCut version (though GrabCut is significantly slower in my testing).

### DIFF
--- a/configs/CMakeLists.txt
+++ b/configs/CMakeLists.txt
@@ -203,6 +203,7 @@ set( CORE_PIPELINE_FILES
   pipelines/common_image_stabilizer.pipe
   pipelines/common_stabilized_iou_tracker.pipe
   pipelines/database_apply_svm_models.pipe
+  pipelines/detector_add_segmentations.pipe
   pipelines/detector_extract_chips.pipe
   pipelines/detector_fish_without_motion.pipe
   pipelines/detector_generic_proposals.pipe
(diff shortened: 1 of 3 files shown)


mattdawkins:
I think we want these to be filters instead of detectors short term, then maybe call them something else

neal-siekierski:
> I think we want these to be filters instead of detectors short term, then maybe call them something else

Agreed that `detector_*` isn't the best prefix. DIVE seems to only show pipelines whose names match `^detector_.+|^tracker_.+|^generate_.+` ([link](https://github.com/Kitware/dive/blob/bad8591471db4dbd03e5583606335d0b1a22ee70/client/platform/desktop/backend/native/common.ts#L192)), so I picked one of those.

PR #127, opened 2021-02-12
Add Wix Installer

mattdawkins:


mattdawkins:
Closing as superseded by the current installer infrastructure. The WiX installer work has since been incorporated and evolved into the current cmake/msi_generate_installer.py workflow and WiX v5 configuration. Rebasing this older WiX v3 implementation would restore outdated infrastructure rather than add a missing installer capability.

PR #137, opened 2021-05-13
Update setup_viame.sh.in

subdavis:


mattdawkins:
probably also needs to be added to setup_viame.install but I can do that at merge time tommorow

mattdawkins (approved):

PR #144, opened 2021-07-21
Check major and minor python versions

russelldj:
This code always failed for me. `PYTHON_VERSION` does not appear to be set until later in the code so I'm checking against the major and minor versions individually. While more verbose, this seems cleaner anyway.


mattdawkins:
This system has been revamped on next

PR #148, opened 2021-10-13
Fix loop foreever in correct_frame_ids_in_viame_csv.py

readicculus:
Fix loop forever.


mattdawkins:
Closing as superseded. The frame-ID utility has moved to examples/detection_file_conversions/standalone_utils/adjust_fid_in_csv_from_list.py, and commit 450b08e06 removed the classifier parsing loop responsible for this hang. Review validation with multiple classes and a trailing note completed successfully and corrected the frame ID. The fix is already present in the current implementation.

PR #152, opened 2022-05-11
Move Addon definitions to CSV file

subdavis:
Several people have asked us how to keep addons in Web up to date.  Until now, we've had to manually enter addon URLs into the update function, but this isn't ideal.

This change will allow VIAME Web (and potentially desktop -- you could imagine a "download addons" button in desktop) to pull the addon manifest directly without requiring any input from the user.  It could also allow users in the future to choose what addons they want to install by name.

The point of this change is to make the addon manifest machine readable.

@mattdawkins will need to update the CMAKE script to make sure this new `download_viame_addons.csv` is included however it needs to be included.

I'd also like feedback from @BryonLewis to make sure this is the best way to accomplish what we need.

### DIFF
--- a/cmake/download_viame_addons.csv
+++ b/cmake/download_viame_addons.csv
@@ -1,0 +1,8 @@
+https://data.kitware.com/api/v1/item/6011e3452fa25629b91ade60/download,habcam
+https://viame.kitware.com/api/v1/item/627b32b1994809b024f207a7/download,SEFSC
+https://viame.kitware.com/api/v1/item/627b3289ea630db5587b577d/download,PengHead
+https://viame.kitware.com/api/v1/item/627b326fea630db5587b577b/download,Motion
+https://viame.kitware.com/api/v1/item/627b326cc4da86e2cd3abb5b/download,EM Tuna
+https://viame.kitware.com/api/v1/item/627b3282c4da86e2cd3abb5d/download,MOUSS Deep 7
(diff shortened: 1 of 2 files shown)


mattdawkins:
Right now add-ons need to be listed in 3 sections: CMakeLists.txt, cmake/download_viame_addons.sh, and configs/CMakeLists.txt so I definitely want to do away with this muliple listings into one file somehow (like this). In addition to the title, including the md5sum as a column entry would be helpful.

The main question is can I get away with removing the links in the README.md and somehow have it use a csv file like this, but still show up as a human clickable link for each add-on in a markdown (or rst file)

subdavis:
Dynamically generating the list in the README would be nice, but it's a bit beyond my capabilities with this particular technology stack.  

Github renders CSV files nicely in the browser: https://github.com/VIAME/VIAME/blob/d2486b71b42b74cada94cf9c2a2bbd091d4c0188/cmake/download_viame_addons.csv

You could just link from the README to this file.  However, the links aren't clickable.

mattdawkins:
That's not bad, I guess we can just link to that from the main page as opposed to having them listed there (I was planning on making that change anyways, though in a different way making a seperate models repository), though we'd probably need to:

(1) Put the title first
(2) add 3 (md5sum) and 4 columns (where 4 is a list of instutitions and authors who contributed to the model, which people have requested be alongside it)

mattdawkins:
The problem with that though is there's windows and linux specific add-ons for some of them, so we'd need an OS column too then consumers of the CSV would need to check it

mattdawkins (approved):

PR #153, opened 2022-06-03
Teach create_mosaic.py a --transparent-black option

predicative:
This PR adds a `--transparent-black` option to the `create_mosaic.py` tool which treats black as transparent when pasting the images.

Note that this is untested.

### DIFF
--- a/tools/create_mosaic.py
+++ b/tools/create_mosaic.py
@@ -95,10 +95,13 @@ def translator(offset):
     result[:-1, -1] = offset
     return result
 
-def paste(dest, src, src_to_dest):
+def paste(dest, src, src_to_dest, transparent_black=None):
     """Copy src into dest, transformed as described by src_to_dest, a
     projective matrix that maps a (2D homogeneous) coordinate in src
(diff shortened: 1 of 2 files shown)


Erotemic:
$0.02: A more general and elegant way to add this feature might be to treat a specific, specifiable pixel value as `NODATA` or `nan`. 

GDAL handles something similar with a `--srcnodata` cli flag: https://gdal.org/programs/gdalwarp.html#cmdoption-gdalwarp-srcnodata or `-n` flags https://gdal.org/programs/gdal_merge.html#cmdoption-gdal_merge-n depending on the program being used.

Here I would recommend using the flag `--nodata` and then accepting either a scalar value that is broadcast to all channels, or a per-channel value.

I have related code in kwplot in the function `crop_border_by_color`
https://gitlab.kitware.com/computer-vision/kwplot/-/blob/main/kwplot/mpl_make.py#L69 that searches for a "nodata" color, but only along the border of the image. Not sure if that's relevant for you here or not.

PR #160, opened 2023-05-18
Few shot detector integration

alexander-lynch:


plugins/pytorch/CMakeLists.txt
```diff
@@ -70,3 +69,29 @@ if( VIAME_ENABLE_PYTORCH-MDNET )
     mdnet_tracker_process )
 
 endif()
+
+if( VIAME_ENABLE_PYTORCH-CUTLER )
```
mattdawkins, line 73: These mmdet* processes are not meant for just running CUTLER and should not be quarantined into requiring it, I haven't looked at the implementation changes of these files yet, but my guess is you're breaking existing functionality in there (the ability to run mmdet models without CUTLER). If the CUTLER option requires a lot of extra dependencies, you can either add a flag in the file itself to enable it's use, or you should make new .py files alongside these specifically for running it and quarantine them off via the ENABLE-CUTLER flag

mattdawkins (changes requested):

.gitmodules
```diff
@@ -70,3 +70,7 @@
 [submodule "packages/pytorch-libs/segment-anything"]
 	path = packages/pytorch-libs/segment-anything
 	url = https://github.com/facebookresearch/segment-anything.git
+[submodule "packages/learn"]
+	path = packages/learn
+	url = https://gitlab.kitware.com/darpa_learn/learn.git
```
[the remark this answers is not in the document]
mattdawkins: Alternatively if you don't want to make that one public due to project considerations we can just clone a version to https://github.com/VIAME (that's what we've done with some other sub-packages) but doesn't matter much either way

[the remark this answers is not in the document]
alexander-lynch: Okay yeah I think that second option would be great. I was trying to find a way to remove the learn dependency entirely by adding in files, but if we can just clone a version that would probably be ideal.

configs/pipelines/detector_convnext_xl.pipe
```diff
@@ -0,0 +1,138 @@
...
+    :chip_height                               800
+    :chip_step_height                          1100
+    :chip_step_width                           600
+    :chip_edge_filter                          -10
+    :batch_size                                1
+
+    block detector:mmdet
+      # Network config, weights, and names
+      relativepath net_config  =               VIAME/src/packages/learn/learn/algorithms/MMDET/configs/convnext_extra_large_config.py
+      relativepath weight_file =               /home/local/KHQ/alexander.lynch/Projects/learn/pretrained_models/cascade_mask_rcnn_convnext_xlarge_22k_3x.pth
```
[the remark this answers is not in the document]
alexander-lynch: On this note where should the pretrained model weights file go?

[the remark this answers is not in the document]
mattdawkins: either data.kitware.com or viame.kitware.com; depending on how large they are we'd either want to put them in a base installer or release as an add-on. I use:

https://viame.kitware.com/girder#collection/5fd27b44c012bf9cfa66fea1/folder/6034155aa267a85e9c328211

Most of the time, and zip up a folder with the models in configs/pipelines/models

Those add-on packages can be used in 3 ways - on the mainpage along installers, in a near future .msi installer for windows as an optional checkmark, and also pulled into a build if set as an enable flag. If setting up the later too and automatic download into VIAME-web as well, you'd want to add them to these two files:

https://github.com/VIAME/VIAME/blob/main/cmake/download_viame_addons.csv

https://github.com/VIAME/VIAME/blob/main/configs/CMakeLists.txt#L122

configs/CMakeLists.txt
```diff
@@ -55,8 +55,23 @@ if( VIAME_DOWNLOAD_MODELS )
         ${VIAME_DOWNLOAD_DIR}/models-mmdet_generic-v1.4.tar.gz
         ${CMAKE_CURRENT_SOURCE_DIR}/pipelines )
     endif()
+    elseif( VIAME_ENABLE_PYTORCH-CUTLER )
+      DownloadAndExtract(
+        https://viame.kitware.com/api/v1/item/646b5af78790390715ac66a7/download
+        ${VIAME_DOWNLOAD_DIR}/VIAME_PoolCar_10_shot_models-V1.0.zip
+        ${CMAKE_CURRENT_SOURCE_DIR}/pipelines )
+    endiff()
```
mattdawkins, line 63: This if statement is malformed, elseif outside of if else block and endiff instead of endif, I fixed it on my push

configs/CMakeLists.txt
```diff
@@ -57,6 +57,22 @@ if( VIAME_DOWNLOAD_MODELS )
...
+        ${VIAME_DOWNLOAD_DIR}/VIAME_PoolCar_10_shot_models-V1.0.zip
+        ${CMAKE_CURRENT_SOURCE_DIR}/pipelines )
+    endif()
+    if( VIAME_ENABLE_PYTORCH-CUTLER )
+      DownloadAndExtract(
+        https://viame.kitware.com/api/v1/item/646b5af78790390715ac66a7/download
+        ${VIAME_DOWNLOAD_DIR}/VIAME_PoolCar_10_shot_models-V1.0.zip
+        ${CMAKE_CURRENT_SOURCE_DIR}/pipelines )
+    endif()
+  endif()
```
mattdawkins, line 74: There's now a number of enable flags added by this branch:

VIAME_ENABLE_LEARN
VIAME_ENABLE_PYTORCH_CUTLER
VIAME_DOWNLOAD_MODELS-POOLCAR
VIAME_DOWNLOAD_MODELS-CONVNEXT

Just for complexity reductions sake, are all necessary? Or can 1-2 be reduced. For example, does having LEARN on but not POOLCAR or COVNEXT make no sense? Alternatively LEARN and CUTLER and they can be consolidated into the same build flag?

Also in the above example, the third if (VIAME_ENABLE_PYTORCH-CUTLER I'm assuming should be poolcar, the second if poolcar changed to convnext, and also the filename for the package download correct

cmake/download_viame_addons.csv
```diff
@@ -6,3 +6,4 @@ EM Tuna,https://viame.kitware.com/api/v1/item/627b326cc4da86e2cd3abb5b/download,
 MOUSS Deep 7,https://viame.kitware.com/api/v1/item/627b3282c4da86e2cd3abb5d/download,MOUSS deep7 fish detection in hawaii
 Aerial Penguin,https://viame.kitware.com/api/v1/item/615bc7aa7e5c13a5bb9af7a7/download,Aerial penguin detection from fixed wings/drones
 Sea Lion,https://viame.kitware.com/api/v1/item/641d644abfe642e7e4f54655/download,Sea lion aerial detection data
+PoolCar,https://viame.kitware.com/api/v1/file/646b5abb4c19222431def881/download, 10 shot Poolcar detector
```
mattdawkins, line 9: The poolcar package here isn't in the same format as the other add-ons, which would be putting any models in a 'config/pipelines/models' folder (this is so that it can be used as an add-on by users downloading it from the installation page, as opposed to just being used in the build system) - I'd probably reccomend changing the package to this format so we can also put them on the install page and here, considering they're 4 Gb. The format that they're in right now would be fine if their a default model in the installer but they won't be due to size.

configs/CMakeLists.txt
alexander-lynch: I ended up combining LEARN and CUTLER into one since they are the same. I think it makes sense keeping the downloads separate as there is a part of the cutler functionality that can be used without the models. I the links and variables should be correct now.

cmake/download_viame_addons.csv
alexander-lynch: I reuploaded the model with that format, so it should be in the right format now

PR #161, opened 2023-05-23
Multicamera Stereo Pairing

Thibault-Pelletier:
Hi,

This MR adds the following stereo camera functionalities to VIAME : 

- Add Stereo Camera calibration process split into calibration checker board corner detection and stereo camera calibration given a series of corner detection
- Add Stereo Disparity Map generation process from undistorted left / right images
- Add Stereo Tracks Pairing process given detected left / right BBox or Polygons and their associated disparity map
- Add 3D center information to left detections in the output VIAME CSV attributes
- Add minor additional arguments for the NetHarn Trainer to specify mini batch size 
- Fix Tain detector train / test split in the context of small datasets where only one image would be taken for the test
- Add specific usage of the previous processes in a specific IFREMER-tk package

@mattdawkins please let me know if you would like me to move any files to the `ifremer-tk` package.
Also, I have left the package as part of this project but I can also move it to another repo if it makes more sense to you.

Thanks in advance for your review,
Best,
Thibault


mattdawkins:
I wasn't a huge fan of adding an packages/ifremer-tk for the following reasons: (1) all other items in the package folder are git submodules of other repos, and (2) all the other repos were heavy code instead of heavy config file, because of this it's closer to an add-on like habcam or seal

On the branch 'next' which I'm currently testing alongside 2-3 other large branches, I refactored the code slightly on commit c95a0e20217fdadfd65aada76791874237e06122, let me know if you agree with the source code moves there. I guess if not, I'd be slightly more partial to making a new repo for this under the VIAME grouping

mattdawkins:
(I can clean up the CMake surrounding that tommorow quick if you agree with it)

Thibault-Pelletier:
Hi @mattdawkins ,

Thanks for the review and your work!
Yes, no problem with moving the code to an addon instead of a package so you can go ahead with the modifications!

Julesdevops:
We're renaming the expected stereo 3d attributes names in dive.  
from `x`, `y, `z` to `stereo3d_x`, `stereo3d_y`, `stereo3d_z`  
I guess this imply some changes here as well when writing pipeline results?

Thibault-Pelletier:
@Julesdevops yes this will impact the following file : 
https://github.com/VIAME/VIAME/pull/161/files#diff-9485b93892a158678dd184a4c861694eac3622ad49df59bc8ab743fb54bf5a14R401

@mattdawkins should I make the modifications or should I let you change it on your side / synchronize with Bryon?

On a related note, it would probably be more interesting to output a 3D BBox instead of just the center point to give more spatial information (fish size approximation).
Not sure what would be the best way to support this (ie. attributes or additional information in the current detection BBox).

Thibault-Pelletier:
@mattdawkins thanks for the merge!

Thibault-Pelletier:
@mattdawkins Thanks again! 

I just created a PR for the stereo name changes here : #162 (@Julesdevops).

For the 3D position, I feel that it should be at the detection level since that's where we extract the information. I'm just concerned that the detection may be clutered at some point with too many detection representations.

Maybe it would be useful to have only one data type stored in a detection (type erased or variant) to support everything and not have to check all the possible types for validity. Access to other types could be left for compatibility and delegated to the one type access (and tagged deprecated).

Having 2D and 3D informations in the object_track_state seems confusing to me as it seems to overlap with the responsibility of the underlying detection state.

Quick design question @mattdawkins and @BryonLewis, I saw in the KWIVER files that we have .proto descriptions for the types. Why don't we use these for communicating between VIAME and DIVE? Is it to allow human manual edition of the CSV files or is it related to something else?

PR #165, opened 2023-09-14
Dev/novelty detection

alexander-lynch:


configs/pipelines/detector_remax_convnext_novelty_detection.pipe
```diff
@@ -0,0 +1,139 @@
...
+    :chip_width                                1333
+    :chip_height                               800
+    :chip_step_height                          1100
+    :chip_step_width                           600
+    :chip_edge_filter                          -10
+    :batch_size                                1
+
+    block detector:detector_remax_convnext
+      # Network config, weights, and names
+      relativepath net_config  =               ../../../../src/packages/learn/learn/algorithms/MMDET/configs/convnext_extra_large_config.py
```
mattdawkins, line 91: Fix paths

configs/pipelines/detector_remax_dino_novelty_detection.pipe
```diff
@@ -0,0 +1,139 @@
...
+    :chip_height                               800
+    :chip_step_height                          1100
+    :chip_step_width                           600
+    :chip_edge_filter                          -10
+    :batch_size                                1
+
+    block detector:detector_remax_dino
+      # Network config, weights, and names
+      relativepath net_config  =               models/DINO_4scale_swin.py
+      relativepath weight_file =               [-MODEL-WEIGHTS-FILE] # ../../../../../pretrained_models/dino/pytorch_dino.pth
```
mattdawkins, line 92: Fix paths

configs/pipelines/train_detector_convnext_remax.viame_csv.conf
```diff
@@ -0,0 +1,71 @@
...
+block detector_trainer:ocv_windowed:trainer
+
+  # Trainer type
+  type = remax_convnext
+
+  # Number of GPUs to use, -1 indicates all
+  remax_convnext:gpu_count = 1
+  remax_convnext:remax_config_file = /home/local/KHQ/alexander.lynch/Projects/learn/VIAME/src/plugins/pytorch/remax/ReMax.yaml
+  remax_convnext:work_dir = /home/local/KHQ/alexander.lynch/Projects/learn/remax_testing
+  remax_convnext:output_directory = /home/local/KHQ/alexander.lynch/Projects/learn/remax_testing
```
mattdawkins, line 65: No absolute paths

configs/pipelines/train_detector_dino_remax.viame_csv.conf
```diff
@@ -0,0 +1,73 @@
...
+
+  # Trainer type
+  type = remax
+
+  # Number of GPUs to use, -1 indicates all
+  remax:gpu_count = 1
+  remax:model_checkpoint_file = /home/local/KHQ/alexander.lynch/Projects/learn/VIAME/src/configs/pipelines/models/pytorch_dino.pth
+  remax:work_dir = /home/local/KHQ/alexander.lynch/Projects/learn/remax_testing
+  remax:output_directory = /home/local/KHQ/alexander.lynch/Projects/learn/remax_testing
+  remax:debug_mode =
```
mattdawkins, line 66: No absolute paths

plugins/pytorch/CMakeLists.txt
```diff
@@ -75,3 +75,19 @@ if( VIAME_ENABLE_LEARN )
   add_subdirectory( cutler )
 
 endif()
+
+if( VIAME_ENABLE_REMAX )
```
mattdawkins, line 79: Where is this option defined? (with a cmake 'option()' command)

If this depends on the learn code, I would just use the VIAME_ENABLE_LEARN flag. If it adds few new dependencies and just depends on pytorch + some new dependencies, then a VIAME_ENABLE_PYTORCH-REMAX advanced flag would probably be best (see root cmake for other pytorch libs and how VIAME_ENABLE_PYTORCH-[SOMETHING] are defined, and also cmake/add_project_pytorch.cmake for how new packages are added). If no new dependencies then probably just use the pytorch flag as opposed to a new one.

plugins/pytorch/remax/model/ops/dist/MultiScaleDeformableAttention-1.0-py3.10-linux-x86_64.egg
mattdawkins, line 1: Binary files should not be checked into the build. Is this an external git subrepo with its own setup.py that should instead be added to packages/pytorch-libs?

mattdawkins (changes requested):
It looks like there's a 3rd party library checked into plugins/pytorch/remax/model/ops    - if this is a 3rd party repo mostly used as-is or with minor modifications, the preferred way to add it to the build would be a git submodule add to packages/pytorch-libs/[module-name] - if there's some changes to the repository, it can be make as a seperate repo under the VIAME group (or another) then that git url added to the superbuild instead of the original repository

configs/pipelines/detector_remax_dino_novelty_detection.pipe
alexander-lynch: What do you mean here? When we added the weights file parameter to the convnext model we used the [-MODEL-WEIGHTS-FILE] placeholder. What should go here instead?

plugins/pytorch/CMakeLists.txt
alexander-lynch: I moved the remax file based on cascade rcnn to the VIAME_ENABLE_LEARN file since it depends on that model. for the other two I added the flag to the highest level CMakeLists.txt file

alexander-lynch:
> It looks like there's a 3rd party library checked into plugins/pytorch/remax/model/ops - if this is a 3rd party repo mostly used as-is or with minor modifications, the preferred way to add it to the build would be a git submodule add to packages/pytorch-libs/[module-name] - if there's some changes to the repository, it can be make as a seperate repo under the VIAME group (or another) then that git url added to the superbuild instead of the original repository

I think it is used without any changes, but it looks like it is only a part of this repo https://github.com/fundamentalvision/Deformable-DETR/tree/main/models/ops is it easier to add the whole repo and then use just the part I want or to make the separate repo with just the parts we need and add that to the superbuild like you said?

configs/pipelines/detector_remax_dino_novelty_detection.pipe
mattdawkins: Pipeline files in the 'configs/pipelines' folder should be default pipelines which work without modification, if this is a template file with values that get filled in by a training routine, it should be in the 'templates' folder not here. Also the comment should be removed #../.../../..   as that has no relevance outside of 1 specific system, and I doubt the find replace string is something like [-MODEL-WEIGHTS-FILE] instead of [-MODEL-WEIGHTS-FILE-] or something symmetrical

plugins/pytorch/CMakeLists.txt
mattdawkins: I would argue again that there shouldn't be a VIAME_ENABLE_REMAX flag, either only a VIAME_ENABLE_PYTORCH-REMAX or not another new ENABLE_X root flag and just re-using the ENABLE_LEARN for this

mattdawkins:
If it's only a subcomponent of a much larger external repo, I could see adding just the subcomponent to either here (or the learn repo) as long as any licenses are preserved and not in violation, and a readme provided in the folder detailing where it came from (and copyrights not violated)

alexander-lynch:
I just checked the licenses of both this and the DINO repo I linked and they are both on Apache 2.0 licenses which should be fine for us. I can add the readme there alongside it.

plugins/pytorch/CMakeLists.txt
alexander-lynch: Okay I'm changing to just use the ENABLE_LEARN flag. There is a use case in which a user wants to use a DINO remax model that they wouldn't need to have LEARN stuff, but that is more a one off example.

configs/pipelines/detector_remax_dino_novelty_detection.pipe
alexander-lynch: I added the direct filepaths, I've been using the config/models/ folder to hold the config and checkpoint files. Is that the best place? It seems to be in the git ignore, so I assume I'll have to add the models somewhere else that will get pulled from.

plugins/pytorch/remax/model/ops/dist/MultiScaleDeformableAttention-1.0-py3.10-linux-x86_64.egg
alexander-lynch: there is a setup.py file in the folder. can that used instead of the github repo?

alexander-lynch:
I added in the readme from the original repo that was used. Is it okay to leave the files where I put them or should I put them in their own repo?

configs/pipelines/detector_remax_convnext_novelty_detection.pipe
mattdawkins: ../../../../src/packages/learn/learn/algorithms/MMDET/configs/convnext_extra_large_config.py is not a valid config path. When running from installers, there is no source directory. All models should be in the models directory underneath these files

configs/pipelines/train_detector_convnext_remax.viame_csv.conf
```diff
@@ -0,0 +1,70 @@
...
+  # Trainer type
+  type = remax_convnext
+
+  # Number of GPUs to use, -1 indicates all
+  remax_convnext:gpu_count = 1
+  relativepath remax_convnext:work_dir = ./
+  relativepath remax_convnext:output_directory = ./
+  remax_convnext:debug_mode = 
+  relativepath remax_convnext:feature_cache = model/convnext_test_features.pt
+  relativepath remax_convnext:net_config  = ../../../../src/packages/learn/learn/algorithms/MMDET/configs/convnext_extra_large_config.py
```
mattdawkins, line 67: fix path

plugins/pytorch/remax/model/ops/MultiScaleDeformableAttention.egg-info/SOURCES.txt
```diff
@@ -0,0 +1,12 @@
+setup.py
+/home/local/KHQ/alexander.lynch/Projects/learn/VIAME/build/install/lib/python3.7/site-packages/kwiver/sprokit/processes/pytorch/utils/model/ops/src/vision.cpp
+/home/local/KHQ/alexander.lynch/Projects/learn/VIAME/build/install/lib/python3.7/site-packages/kwiver/sprokit/processes/pytorch/utils/model/ops/src/cpu/ms_deform_attn_cpu.cpp
+/home/local/KHQ/alexander.lynch/Projects/learn/VIAME/build/install/lib/python3.7/site-packages/kwiver/sprokit/processes/pytorch/utils/model/ops/src/cuda/ms_deform_attn_cuda.cu
```
mattdawkins, line 4: Fix paths

mattdawkins (changes requested):

mattdawkins:
The original repository just with whatever changes (including reductions if it's only a small part of the repository) were made on top is much more ideal then copy and pasting them into VIAME as-is. That way (a) the history of the original formation of the files is preserved and (b) if external 3rd party changes are made to the same files (e.g. package updates) they can be easily merged into the code.

That said, if it's too long to do I can merge it with copy pasted as long as there's a readme and license there, it's just less ideal (I did it in one instance for example, but didn't in five others).

alexander-lynch:
I currently added everything in line with leaving the copy pasted repo in there just to have something that I think will work. I'm going to go back now and try and do it with the github repo way to see if I can get that working instead

PR #178, opened 2024-07-04
[fix] Fix issues when building on windows 10 and/or python 3.10

medmssaouri:
Goal : 
  - Fix bugs when building on windows 10 or/and python 3.10, and add windows/ubuntu workflows.

Changes : 
- Fix:
   - Disable Bezier when the Python version is greater than or equal to 3.9 because version 2020.1.14 only supports up to Python 3.8. Upgrading Bezier to a version that supports Python 3.10 will require upgrading numpy, which in turn will necessitate a  significant number of modifications to other Python/PyTorch packages.   
   - Upgrade the default version of FFmpeg to 4.4.1 in windows because version 3.3.3 causes issues in the build of VIAME/fletch  on Windows 10. 
   - Upgrade ubelt to version 1.3.3 due to the new version of liberator, which is installed automatically by torch_liberator, itself installed automatically by bioharn.
   - Customize PYTHONHOME on windows only when VIAME_ENABLE_PYTHON-INTERNAL is enabled because of this issue: [encodings](https://stackoverflow.com/questions/65184937/fatal-python-error-init-fs-encoding-failed-to-get-the-python-codec-of-the-file)
   - Add ubuntu 20.04 build workflow.
   - Add Windows 10 build workflow.

The changed above are tested to build successfully on windows 10 and ubuntu 22.
 The changed above are tested on inference of yolo-darknet and crcnn.


medmssaouri:
@ltetrel Could you test this branch from your side

mattdawkins:
What's the motivation of updating the fletch ffmpeg links? I think they're still valid for me

medmssaouri:
> What's the motivation of updating the fletch ffmpeg links? I think they're still valid for me

@mattdawkins Yes, the FFmpeg links are still working, but I think that having links pointing to data.kitware is safer in terms of availability. I can drop that commit if you prefer to keep them pointing to their original links.

cmake/add_project_fletch.cmake
```diff
@@ -199,6 +199,11 @@ if( VIAME_ENABLE_FFMPEG )
       ${FLETCH_DEP_FLAGS}
       -DFFmpeg_SELECT_VERSION:STRING=2.6.2
     )
+  elseif( WIN32 )
+    set( FLETCH_DEP_FLAGS
+      ${FLETCH_DEP_FLAGS}
+      -DFFmpeg_SELECT_VERSION:STRING=4.4.1
```
mattdawkins, line 205: Does the VXL ffmpeg plugin still get built with this change?

If you want to proceed with 4.4.1 but I want to continue with 3.3 for now we might just need to parametize this variable (VIAME_FFMPEG_VERSION or the like HL build flag)

cmake/add_project_python_deps.cmake
```diff
@@ -118,9 +118,13 @@ if( VIAME_ENABLE_PYTORCH-NETHARN AND VIAME_ENABLE_GDAL )
 endif()
 
 if( VIAME_ENABLE_OPENCV OR VIAME_ENABLE_PYTORCH-NETHARN )
-  list( APPEND VIAME_PYTHON_BASIC_DEPS "ubelt<=1.2.3" "pygments" "bezier==2020.1.14" )
+  list( APPEND VIAME_PYTHON_BASIC_DEPS "ubelt==1.3.3" "pygments")
```
mattdawkins, line 121: Not sure if this will cause issues on Python 3.6, might need to select version based on python version but I can go back and do later if needed after attempted build with change

Erotemic: I'm not quite sure how `VIAME_PYTHON_BASIC_DEPS` are being used, but if you are passing them to pip you can use environment markers so you don't have to think about version selection in CMake. E.g. In a Python 3.11 environment, If I run:

```bash
pip install "coverage>=6.1.1; python_version< '3.9' and python_version >= '3.8'"
```

It does nothing and prints the message:

```
Ignoring coverage: markers 'python_version < "3.9" and python_version >= "3.8"' don't match your environment
```


But it will install the package if I run:

```bash
pip install "coverage>=6.1.1; python_version < '3.12' and python_version >= '3.10'"    
```

However, ubelt up to version 1.3.6 still support Python 3.6, although I do plan to drop support for 3.6 and 3.7 in the next version.

cmake/add_project_fletch.cmake
medmssaouri: In my build VXL and ffmpeg are activated, and i can viame_processes_vxl.dll, kwiver_algo_vxl.dll, and kwiver_algo_ffmpeg.dll ..., is that what you mean by "VXL ffmpeg plugin"

mattdawkins: Not exactly - VXL has it's own ffmpeg reader, which would be in the kwiver_algo_vxl.dll plugin. It gets built whenever VXL is enabled, but if VXL isn't build with ffmpeg support, it will just error out at runtime instead of build time. This has been VIAME's default video reader for a while, but we should switch to the direct ffmpeg wrapper after confirming that there are no regressions with it, or modifications to the frames output from it due to the potential for the frames being slightly temporally off between different readers.

https://github.com/Kitware/kwiver/blob/master/arrows/vxl/vidl_ffmpeg_video_input.h

PR #181, opened 2024-10-22
refactor: split logic into functional parts

Erotemic:
Refactor to make the logic more clear.

* Splits the single global-scope main into smaller functions

* The annotations mapping from a List to a List of Lists which has a specific structure now uses a List of dictionaries to make the structure more clear from reading the code.


mattdawkins:
This has been merged into next and built upon

mattdawkins (approved):

PR #182, opened 2024-11-18
Add json output option to calibrate_cameras.py (stereo support)

collinsr:
Add -j / --json option to tools/calibrate_cameras.py. The resulting json file can be loaded by kwiver's vital/io/camera_rig_io code.


collinsr:
@mattdawkins

PR #186, opened 2024-11-30
feat: add sam2 refiner

Erotemic:
Hooks into the `sam2` module to refine a VITAL `DetectedObjectSet`. Currently uses the bounding boxes as SAM2 prompts, gets the masks from SAM2, and then converts the masks back into what VITAL expects (masks relative to the bbox positions).

Here is the inital result for the example in the docstring:

sam2_refined

Because only the mask is updated (and note the original mask is not used as a prompt, although it could be), the mask for the astronaut has a lot of holes and is speckley in places.

### DIFF
--- a/plugins/pytorch/sam2_refiner.py
+++ b/plugins/pytorch/sam2_refiner.py
@@ -1,0 +1,517 @@
+# ckwg +29
+# Copyright 2024 by Kitware, Inc.
+# All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


Erotemic:
I've added two configuration options:

* hole_policy - Can be allow or discard
* multipolygon_policy - Can be allow, convex_hull, or largest

The following is with hole_policy=discard, multipolygon_policy=largest

sam2_refined

PR #195, opened 2025-02-17
Add win32 libs to pythonpath on Windows

Louis-Pagnier-KW:
SAM2 on Windows requires some libraries (such as pywin32, pywintypes, ...) that are installed in subdirectories of site-packages and are not findable with the current PYTHONPATH.

Otherwise, we encounter the following error:
`Caught unhandled std::exception: InstantiationException: Error locating target 'sam2.modeling.backbones.hieradet.Hiera', set env var HYDRA_FULL_ERROR=1 to see chained exception.`


mattdawkins:
While it probably doesn't matter too much, I'm not sure if adding special cases to the path is better than just modifying/fixing where these libraries get installed to in the first place - either within the library installing them (if fast) or just as a cmake patch after the library is installed moving the file

Louis-Pagnier-KW:
Just moved to the cmake patch solution.
I Agree with you, it's better than having a giant python path.

mattdawkins:
Sorry, got side-tracked on other projects for a bit, but I'm coming back to finishing off these installers this week (probably into next because there's still a few outstanding issues, though only with auxiliary functionality). This script wouldn't work for a few reasons:

(1) It's only called when installing torch from pip, not when using the internal build

https://github.com/VIAME/VIAME/blob/8ea541a8540b146bb57c52a70575bab1a402a83b/cmake/add_project_python_deps.cmake#L290

(2) It occurs before SAM2 is built or installed, directly after torch is built, therefore it wouldn't be run unless it's a 2nd build not a first

(3) The enable SAM variable won't be passed down into it since running a cmake patch in this context doesn't inherit all environmental variables from the parent, only what's defined in the above passdown

Lastly, I'd recommend a move not a copy so the binary folder isn't duplicated twice, assuming that still works in the installer

I'd probably just recommend adding it to the below for now:

https://github.com/VIAME/VIAME/blob/next/cmake/custom_install_viame.cmake#L3
https://github.com/VIAME/VIAME/blob/8ea541a8540b146bb57c52a70575bab1a402a83b/cmake/add_project_viame.cmake#L79

Which gets called at the end of the installation process, though alternatively you could add a custom install rule for the SAM2 dep in:

https://github.com/VIAME/VIAME/blob/next/cmake/add_project_pytorch.cmake#L190

Let me know if you want to revise it, if not I could write a script for it tommorow. 

Only semi-related, but looking at the custom_torch script, I can just remove it from the 'next' branch since it's meant for pytorch 1.0 and is now deprecated

Louis-Pagnier-KW:
Thank you !

I tried to add the custom installation to the `custom_install_viame.cmake` file but I currently cannot test it as I don't have a Windows machine.

Also, as `win32/lib/` is contained in `win32/`, I move `win32/lib/*` first, delete it, then move `win32/*`.

mattdawkins (approved):

PR #201, opened 2025-03-21
Feat/huggingface zeroshot detector

Erotemic:
Adds a new object detection algorithm that uses hugging face GroundingDino object detection for zero-shot detection. By default the prompt is "foreground object", which should give reasonable results. Specifying what you are looking for exactly should do even better (so long as it is a natural image sort of thing).

https://huggingface.co/docs/transformers/model_doc/grounding-dino


Erotemic:
I think this is now in working order. I've been able to run a demo. 

Setup:

```bash
cd /opt/noaa/viame/
export PYTHONIOENCODING=utf-8
source setup_viame.sh

# Test a Huggingface detector
cd /opt/noaa/viame/examples/object_detection
cp /opt/noaa/viame/configs/pipelines/detector_huggingface_zeroshot.pipe detector_huggingface_zeroshot.pipe
```

Running a generic detector and a detector with a specific prompt:
```
# Run the generic detector
kwiver runner detector_huggingface_zeroshot.pipe \
              -s input:video_filename=input_image_list_small_set.txt \
              -s detector_writer:file_name=generic_detections.json \
              -s detector_writer:writer:type=coco
```

This results in the following detections on the example images:

stack_1c2d164236579e75

You can also run with a specific prompt:

```
# Use a specific prompt
kwiver runner detector_huggingface_zeroshot.pipe \
              -s input:video_filename=input_image_list_small_set.txt \
              -s detector1:detector:ocv_windowed:detector:huggingface_zeroshot_detector:classes="[fish,starfish,coral]" \
              -s detector_writer:file_name=prompted_detections.json \
              -s detector_writer:writer:type=coco  
 ```

stack_664f8da0ac2e8f78


While doing this I noticed that the coco files are output with some strange combinations of classes. I'm not sure exactly where that is coming from. In the prompted case, I get how it might combine two overlapping detections with different classes, but in the first case there are two categories: "foreground object", and "fore object", so I don't quite understand how that's happening. I suspect it is a bug.

I also noticed the coco files have the very very old (and misspelled) auxillary field with empty channels:

```
{
    "id": 9,
    "file_name": "201503.20150528.090713303.292184.png",
    "auxillary": [
        {
            "file_name": "201503.20150528.090713303.292184.png",
            "channels": ""
        }
    ]
}
```

In the case where there is only a single image, that should be omitted, and if the channels aren't known they should not be specified.

Erotemic:
Probably needs to be reopened.

PR #211, opened 2026-01-23
Resolves #210

### LINKED ISSUE #210: avoid writing config to disk for yolo-mit
In our public instance of viame-web, the installation path is read-only. But because the yolo-mit integration currently writes a dataset config for training into the installation folder (for unnecessary reason), we are not able to run a yolo training.

The solution is either for us to use a tmp dir, or better to use directly the yolo-mit python api to avoid unnecessary disk write.

ltetrel:
Solution is to use the yolo-mit python api instead of doing a system call (which is a really bad behavior IMO).


Erotemic (approved):
I'm going to give this a tentative approval. I think it is an improvement.

Have you verified that this works to train a model? If so I think its safe enough to merge.

PR #236, opened 2026-04-07
fix: Removed 'timestamp' from detector writer of tracks_only_adjust_csv

### LINKED ISSUE #235: Tracks only adjust CSV pipeline crash
The pipeline crash when running on detections.
An extra connection to non-existing `timestamp` port of `detector writer` process.

Note: This port exist on the similar `track_writer` process.

Louis-Pagnier-KW:
Closing #235 

The detector_writer process doesn't have any `timestamp` port, but a connection to it is declared in `filter_tracks_only_adjust_csv.pipe` file, leading to a crash.

### DIFF
--- a/configs/pipelines/filter_tracks_only_adjust_csv.pipe
+++ b/configs/pipelines/filter_tracks_only_adjust_csv.pipe
@@ -97,8 +97,6 @@ process detector_writer
 
 connect from downsampler2.output_3
         to   detector_writer.detected_object_set
-connect from downsampler2.timestamp
-        to   detector_writer.timestamp
 connect from image_writer.image_file_name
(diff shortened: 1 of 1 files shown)


mattdawkins:
Closing as obsolete on current main. The original diagnosis applied to an older KWIVER writer, but `detected_object_output` now declares an optional `timestamp` input and consumes it when connected. Removing the connection is therefore no longer needed to prevent an unknown-port crash.

During the review, both the original `filter_tracks_only_adjust_csv.pipe` and the version with this connection removed completed successfully on a synthetic one-frame input and produced the same detection rows.

This PR has the same patch as #266; both are being closed as obsolete. Neither fix is being merged.

PR #240, opened 2026-04-10
fix: set uint16_scaled as default format

### LINKED ISSUE #239: Set uint16_scaled output format as default for disparity map
Currently, the output_format of the `measurement_compute_rectified_disparity` pipeline is `raw`.  
The problem with this format is that if a disparity is higher than 255, kwiver will cap it to 255.

So if the defined unit is small (e.g. millimeters), we can quickly have values above 255 and have a white image with lost information.

Louis-Pagnier-KW:
Closing #239


mattdawkins:
Closing as already implemented. The current measurement pipeline on main already uses uint16_scaled for disparity output with the corresponding scale of 256. This PR would not add a remaining behavior change to the current pipeline.

PR #248, opened 2026-04-27
Add fast foundation stereo and upgrade onnx

mattdawkins:


cmake/add_project_onnx.cmake
```diff
@@ -1,6 +1,10 @@
-# Install onnx python
-set( ONNXRUNTIME_PYTHON onnxruntime )
-set( ONNXRUNTIME_VERSION 1.12.1 )
+# Install onnx python. onnxruntime-gpu (vs vanilla onnxruntime) gives the
+# CUDAExecutionProvider that fast_foundation_stereo_onnx and similar GPU
+# inference wrappers need; it imports as the same `onnxruntime` package
+# so existing CPU-path consumers keep working. Version 1.23.x is the
+# current line that supports CUDA 12 (matches our torch wheel).
+set( ONNXRUNTIME_PYTHON onnxruntime-gpu )
+set( ONNXRUNTIME_VERSION 1.23.2 )
```
mattdawkins, line 7: @Louis-Pagnier-KW 

Is there any reason why you need to keep the onnx version lower here?

Louis-Pagnier-KW: @ltetrel do you know if there is any reason of why it was fixed at 1.12.1 originally ?

ltetrel: It was important on our side to keep compatibility with old ONNX so it works with old jetson devices for our client. Is there any reason to upgrade ?

ltetrel: @mattdawkins Please ask us before merging anything related to ONNX, it is a critical path for us...

mattdawkins: Yeah, I didn't mean to merge this one which is why I put it as a PR in the first place, new PR:

https://github.com/VIAME/VIAME/pull/249

PR #249, opened 2026-04-28
Add fast foundation stereo and update python ONNX

mattdawkins:


mattdawkins:
@ltetrel I'd prefer not to keep an older CPU-only onnx version from 2022 as the default. ONNX runtime >= 1.17 is required for CUDA 12+ support, which is currently shipped in the binaries.

Maybe this should be parametized however, so that you can still do older builds against an older ONNX? Is your user using the default desktop binaries or a custom build?

ltetrel:
> @ltetrel I'd prefer not to keep an older CPU-only onnx version from 2022 as the default. ONNX runtime >= 1.17 is required for CUDA 12+ support, which is currently shipped in the binaries.
> 
> Maybe this should be parametized however, so that you can still do older builds against an older ONNX? Is your user using the default desktop binaries or a custom build?

Yes, it should be parametrized. I will check locally if at least we have some regressions when upgrading, can I use that branch for testing ?

mattdawkins:
Right now this branch just makes the python package be the updated version, but leaves the C++ version used in mmdeploy unchanged (I'd like to update them both, but if this works as an intermediate option, it might work shorter term)

ltetrel:
> Right now this branch just makes the python package be the updated version, but leaves the C++ version used in mmdeploy unchanged (I'd like to update them both, but if this works as an intermediate option, it might work shorter term)

Ok so can I use that branch for testing ?

mattdawkins:
Branch for testing as in a compiled version of the code? There's a windows version here:

https://drive.google.com/file/d/1s2KMNC13NRiLmQnZ8E5WIhFStfScbVvg/view?usp=drive_link

If just looking for a source branch, 'main' lacks this change, but this branch 'dev/add-fast-fdn-stereo' has it.

cmake/add_project_python_deps.cmake
```diff
@@ -157,6 +157,11 @@ endif()
 if( VIAME_ENABLE_ONNX )
   list( APPEND VIAME_PYTHON_BASIC_DEPS "onnx<=1.18")  #keep old compatibility for TensorRT<10 https://onnxruntime.ai/docs/execution-providers/TensorRT-ExecutionProvider.html#requirements
   list( APPEND VIAME_PYTHON_BASIC_DEPS "onnxscript" )  #https://github.com/pytorch/pytorch/issues/166352
+  if( VIAME_ENABLE_CUDA )
+    list( APPEND VIAME_PYTHON_BASIC_DEPS "onnxruntime-gpu==1.23.2" )
+  else()
+    list( APPEND VIAME_PYTHON_BASIC_DEPS "onnxruntime==1.23.2" )
+  endif()
```
ltetrel, line 164: `onnxruntime-gpu` is not specific to CUDA, but it is okay to keep it here

plugins/pytorch/CMakeLists.txt
```diff
@@ -284,6 +284,16 @@ if( VIAME_ENABLE_PYTORCH-STEREO )
...
 
+    kwiver_add_python_module(
+      ${CMAKE_CURRENT_SOURCE_DIR}/fast_foundation_stereo.py
+      pytorch
+      fast_foundation_stereo )
+
+    kwiver_add_python_module(
+      ${CMAKE_CURRENT_SOURCE_DIR}/fast_foundation_stereo_onnx.py
+      pytorch
+      fast_foundation_stereo_onnx )
```
ltetrel, line 295: Why making two separate modules for ONNX version and pure python counterpart ?

packages/pytorch-libs/fast-foundation-stereo
ltetrel, line 1: There are a lot of submodules, which complexify code maintenance. Can't we just add the `fast_foundation_stereo` library as a cmake/python dependency alone without forking ?
(this also applies for many others plugins)

ltetrel (commented):
Some comments around plugin architecture.

ltetrel:
@mattdawkins  I was also able to test the export after your ONNX upgrade to 1.23.2, and there was no regressions :heavy_check_mark:.

packages/pytorch-libs/fast-foundation-stereo
mattdawkins: A lot of the python submodules added aren't on pypip or require matched cuda versions so building them makes sense. One of the motivations behind the git submodules is being able to edit them quickly when needed as well, for the newer faster moving repos that might need edits.

Because many are optional, I could see doing what we did for the LEARN repository so they're not un-necessarily downloaded when not needed by users. The LEARN repo was so big I wrote a quick script that only checks out the repos when needed but are still placed in the packages directory.

plugins/pytorch/CMakeLists.txt
mattdawkins: One is pure torch the other onnx, they should probably be selectively enabled and I'm going to move the onnx one to the newer onnx plugin dir

cmake/add_project_python_deps.cmake
mattdawkins: Are you sure? I thought AMD and intel required different optimized runtimes / package names (e.g. onnxruntime-directml)

PR #250, opened 2026-04-28
Feat: Add Dive high level parameters metadata

Louis-Pagnier-KW:
This PR add meta data to pipeline files so Dive can show a popup to customize pipeline runs.

Currently, I made these metadata directly in pipeline files.  
The main advantage is that it is easier to maintain, it is harder to forgot to update metadata as it is in the same file on the same line.

However, if later we decide to migrate to dedicated files, we can make a script to automatically extract these metadata from pipe files to a new format.

Also, thanks to customization, I was able to remove some pipelines that were almost the same, but with few different settings: (e.g. empty frame lbls Xfr, ...).

The Dive part of this PR is already merged and included in DIVE 1.9.11:
https://github.com/Kitware/dive/pull/1624

So, this MR also update DIVE from 1.9.10 to 1.9.11.


mattdawkins:
I have some minor issues with the branch, but I think it's okay as a first pass and could be merged - except for the DIVE non-binary hash change (particularly packages/dive, not the binary links themselves those are fine - though I could always revert that out after merge).

1) All of the parameter types (bool, int) is already defined in the C++ code so later on I think we should just export this data from the kwiver plugin loader or some other process into a json or something eventually (not on this branch). Putting these params after the definition in the pipeline, but on the same line, makes the .pipe files super wide as well, whereas previously I tried to keep them to some consistent character limit for many of the pipelines.

2) Some of the INPUT and OUTPUT specifications don't cover everything the process does, do they support multiple listings for that output field? MASK and BOX or the like - I might want to iterate more on these later

3) On main, I had a concept of required parameters that had to be filled in before pipeline run that I need for some quick stereo stuff that's specified in a slightly different way in the initial comments. A DIVE branch for that is up already for this style of params - technically that branch would need to land too to prevent regression

4) This has to do with your DIVE branch, not this one, but putting icons next to the pipeline name to the right of them was confusing in a few instances. One looks like a right arrow, but then when you scroll over it the tool info opens on the left side. I think this icon either needs to be changed or removed.

ltetrel:
>    1. All of the parameter types (bool, int) is already defined in the C++ code so later on I think we should just export this data from the kwiver plugin loader or some other process into a json or something eventually (not on this branch). Putting these params after the definition in the pipeline, but on the same line, makes the .pipe files super wide as well, whereas previously I tried to keep them to some consistent character limit for many of the pipelines.

We also had this conversation itnernally, duplicating the type from c++ is indeed not a good choice. What process were you thinking when you say "pulling types from kwiver" ? Some process at runtime that generates a json on-the-fly ? A mechanism in DIVE to parse that type from cpp source code directly ?

>    2. Some of the INPUT and OUTPUT specifications don't cover everything the process does, do they support multiple listings for that output field? MASK and BOX or the like - I might want to iterate more on these later

Same as your previous comment, this could be parsed from cpp automatically.

>    3. On main, I had a concept of required parameters that had to be filled in before pipeline run that I need for some quick stereo stuff that's specified in a slightly different way in the initial comments. A DIVE branch for that is up already for this style of params - technically that branch would need to land too to prevent regression

I am not sure to follow here, you mean you already set-up a similar thing that @Louis-Pagnier-KW proposed here ? If so can you point us to an example.

Overall, even though this is definitively a great and necessary addition on the DIVE side (for user transparency), I have doubt about this design, why this lives on the kwiver side? Why a kwiver pipeline should have something specific to DIVE ? I have similar concerns as @mattdawkins, that those changes make the overall pipeline file un-readable at some points, and particularly hard and cumbersome to develop and maintain.

Louis-Pagnier-KW:
If extract c++ pipelines params automatically to a JSON file, I think the script should only change base type but still allow manual control.

For example, running the script for the first time would output something like that:

```json
{
  "params": {
    "iou_treshold": {
      "label": "Iou Threshold",
      "path": "my_process:my_block:iou_threshold",
      "type": "double",
      "min": null,
      "max": null,
      "editable": false
    },
    "weird_internal_param": {
      "label": "Weird Internal Param",
      "path": "my_process:my_block:weird_internal_param",
      "type": "unsigned int",
      "min": null,
      "max": null,
      "editable": false
    }
  }
}
```

Once, generated we would have to update manually some extra config, such as automatically generated labels or additional type restrictions, such as 0-1 limitation for ratio, or non-zero values for some parameters.

Manual edition:
```json
{
  "params": {
    "iou_treshold": {
      "label": "IoU Threshold", // <--- fixed 'U' of auto generated label
      "path": "my_process:my_block:iou_threshold",
      "type": "double",
      "min": 0, // <--- Add min limit that make sense
      "max": 1, // <--- Add max limit that make sense
      "editable": true // <--- Enabled the param because it is important to be editable
    },
    "weird_internal_param": {
      "label": "Weird Internal Param",
      "path": "my_process:my_block:weird_internal_param",
      "type": "unsigned int",
      "min": null,
      "max": null,
      "editable": false // <--- We don't want users to change this param
    }
  }
}
```

Then, if later we run again the script to extract params, it should add new params, remove those that are removed, but, I think, only change type if modified, but keep manual added constraints:

Re-run the script with some changes to c++ plugin:
```json
{
  "params": {
    "iou_treshold": {
      "label": "IoU Threshold", // <--- Keeping manual label
      "path": "my_process:my_block:iou_threshold",
      "type": "float", // <--- Auto replacing type
      "min": 0, // <--- Keeping manual min
      "max": 1, // <--- Keeping manual max
      "editable": true // <--- // <--- Keeping it editable
    },
    "weird_internal_param": {
      "label": "Weird Internal Param",
      "path": "my_process:my_block:weird_internal_param",
      "type": "unsigned int",
      "min": null,
      "max": null,
      "editable": false
    }
  }
}
```

They are a lot of params in many pipelines that we don't especially want users to edit (too complex or required for the pipeline to work properly in DIVE context).
So, I think by default, they could all be disabled by default, and then we can quickly pass on them to enable those that make sense.

For python, pipelines, type could be extracted from type hints, but are not guaranteed.

What do you think of this ?

We could do it in a future MR to replace the DIVE_PARAMS system in `.pipe files.

Also, a script can be written to edit JSON generated files based on DIVE_PARAMS of `.pipe` files based on param path, then remove them. So, we won't loose all that work.

ltetrel:
Seems much better yes, do you think that this could be run-able at runtime by DIVE, or we want to serialize that file to disk @mattdawkins  ? Maybe we should get input from @BryonLewis  as well on this.

ltetrel:
I think it is also important to look around and see how things are done for our other projects. For e.g. ParaView uses [XML spec files](https://www.paraview.org/Wiki/ParaView/Plugin_HowTo#XML_Only) to define the UI, whenever you create a filter, you need to define all your parameters. It is not automatically parsed from source code sadly. We can have something similar in json.
Here I would go with YAML or TOML files for even better clarity (we don't care of performance of XML).

mattdawkins:
>> I am not sure to follow here, you mean you already set-up a similar thing that @Louis-Pagnier-KW proposed here ? If so can you point us to an example.

Only for a smaller number of config values, but here:

https://github.com/VIAME/VIAME/blob/main/configs/pipelines/measurement_calibrate_cameras_fast.pipe#L6

Once the 'required' specification lands in this DIVE for this format, we can replace those however, so it's not duplicated, then once the segmentation branch lands in DIVE put the packages/dive hash back to main

PR #251, opened 2026-04-30
Pipeline tests

Louis-Pagnier-KW:
Tests for every pipeline DIVE can run, under `tests/pipelines/`, exercised the way DIVE runs them (same inputs, same output paths).

### Layout

```
tests/pipelines/
├── cases.py          # discovery + per-family rules + per-pipeline overrides
├── conftest.py       # input fixtures (ENVIRONMENTS table) and ctest skip mapping
├── validators.py     # check_csv, check_generated_frames/chips/video
├── viame_runner.py   # kwiver runner wrapper with a per-pipeline timeout
├── test_<category>.py × 6   # one parametrize over discover(<category>)
└── pipelines_test_data/     # downloaded by CMake
```

### How pipelines are found

`discover(category)` globs `$VIAME_INSTALL/configs/pipelines/<category>_*.pipe` and applies the same exclusions as `dive_tasks.pipeline_discovery`. Each match gets:

- the category default (fixture + params + check), then
- every matching `RULES` regex in order (a family like `detector_sea_lion_*` shares one fixture and assertion), then
- an `OVERRIDES` entry if one exists (per-pipeline config, fixture, or a custom check; a list expands into several cases over one pipeline).

A pipeline matching nothing is a skip reading `no test case defined`, so new pipelines surface instead of going unnoticed. Add-on pipelines are covered without an edit here.

### Assertions

`check_csv` (count with equal/min/max, all-`polygon` / all-`head-tail`, stereo pairs), `check_generated_frames`, `check_generated_chips`, `check_generated_video`, plus custom callables for KWA, homographies, depth maps and calibration output.

### ctest

Six entries, one per category, label `PIPELINES`:

```
ctest -L PIPELINES
```

Single pipeline: `pytest tests/pipelines/test_detector.py -k detector_default_fish`.

Granularity is per category rather than per pipeline: add-on pipelines only exist after the build, so CMake cannot enumerate them at configure time.

### Behaviour worth knowing

- Each pipeline run has a 900s timeout; a hung pipeline fails with its name instead of consuming the category's 7200s budget.
- A missing install or `configs/pipelines` directory is a collection error, not a skip.
- ctest sees exit 5 (skipped) only when nothing passed and nothing failed.
- Two pipelines fixed along the way: `transcode_tracks_only` (missing connect `to`) and `utility_add_head_tail_keypoints_from_dets` (conflicting `mask_to_poly_*` settings).


mattdawkins:
Some minor consolidation with existing tests, which may or may not help further:

https://github.com/VIAME/VIAME/pull/258

Louis-Pagnier-KW:
@mattdawkins, can we merge this ?

mattdawkins:
Going to look at it some more next day / two just to make sure there's no duplication with existing tests

mattdawkins:
My other main concern was that there's a lot of tests pointed at individual pipelines, and sometimes pipelines get renamed/deleted a bit - having the update these tests for each seems cumbersome as opposed to looping over all pipes in some directory with arbitrary filenames, other

ltetrel:
> My other main concern was that there's a lot of tests pointed at individual pipelines, and sometimes pipelines get renamed/deleted a bit - having the update these tests for each seems cumbersome as opposed to looping over all pipes in some directory with arbitrary filenames, other

Regarding your concern over maintainability and making the discovery automatic, I believe that fixing tests is the classical way to go (and we agreed on that with @Louis-Pagnier-KW when drafting the implementation). Unit tests usually relies on fixtures (for e.g. [python fixture](https://docs.pytest.org/en/stable/explanation/fixtures.html)), with one test = one feature. In VIAME, one test = one config.
With your test design, if a pipeline file is deleted or renamed, a dynamic loop would simply skip it without throwing an error.
Hardcoding these test cases ensures that any structural changes are explicitly caught.

mattdawkins:
I still think I'd prefer globing through each category of pipeline for adding tests that don't have any specialized pipeline IO checks (specific to that pipeline), but retaining the option to over-ride this and provide specialized IO/runtime checking or parameters for specific pipelines.

I did this on the last few commits, and think this is ready to merge if you don't object. 

>with one test = one feature

This doesn't change this, there's still one test per pipeline they're just not explicitly referenced in the pipeline code by name, unless needed for a custom test

PR #258, opened 2026-06-22
Dedup test infrastructure: shared install/env helpers + single viame_add_pytest_test

mattdawkins:
Follow-up to #pipeline-tests: removes duplicated test plumbing introduced alongside the new `tests/pipelines/` tree.

### Refactor
- New `tests/common/`, shared by examples, pipelines, and the pytorch/SAM3 plugin tests:
  - `viame_env.py` — VIAME install discovery + `setup_viame.sh` sourcing (previously duplicated **4 ways**: `viame_runner.py`, `examples/test_utilities.py`, and `plugins/pytorch/test_sam3.py`).
  - `cmake/add_pytest_test.cmake` — a single `viame_add_pytest_test()` registering pytest classes as ctest tests. A `SOURCE_SETUP` option covers the plugin tests that must source `setup_viame.sh` (native torch/kwiver imports); examples/pipelines use the default PYTHONPATH/`VIAME_INSTALL` mode.
- Hoisted `viame_source`/`viame_install` fixtures to `tests/conftest.py` (self-paths `tests/common`, so it's safe for every subtree).
- Per-subtree CMakeLists keep their thin wrappers, so all per-test call sites are unchanged.

### Bug fixes
- `env_fish_with_detections` copied `fish_1_jpg` (missing dot) → `fish_1.jpg`.
- `validators.check_csv` stereo branch used `name.replace("1","2")` (rewrites any `1`) → swap only the trailing camera index.

### Verification
- `py_compile` on all touched Python; import wiring exercised via a pytest-stub harness.
- cmake parses; stubbed `add_test`/`set_tests_properties` output matches the originals for all three subtrees (default + `SOURCE_SETUP`).
- Not run: full `ctest` (needs a built VIAME install).


Louis-Pagnier-KW:
Thanks for this refactor!
I just fixed few things related to cmake that were crashing my build.

Tested and works for pipelines tests + example tests

I think you can merge this branch in `pipeline-tests`

PR #267, opened 2026-07-03
fix: mit-yolo YAML now correctly uses categories from kwcoco instead …

### LINKED ISSUE #260: Possible label mismtach with yolov9
Ifremer reported some possible mismatch between labels. I am still waiting on their dataset to reproduce the error, it may come from their taxonomy as the species may not be sufficiently independent and are too closely related. If that is the case I will close that issue.
In the meantime, I will train / evaluate on COCO and see what happens.

ltetrel:
…of kwiver CategoryHierarchy

The proposal fix implements in `_ensure_format_writers`, ideally we could fix directly in kwiver, but I am not even sure if that make sense. After all, the `labels.txt` maping is completely independent from kwcoco catgeories, just category names should stick. Anyway it can be confusing for the developper (and it happened to me) that the atribute `_categories` is the one from kwiver, and not kwcoco.

### DIFF
--- a/plugins/pytorch/mit_yolo_trainer.py
+++ b/plugins/pytorch/mit_yolo_trainer.py
@@ -164,15 +164,25 @@ class MITYoloTrainer( KWCocoTrainDetector ):
             self._training_writer.complete()
             self._validation_writer.complete()
 
-            # hack, need to fixup the writers
+            # conforming kwcoco train and val datasets and set categories to the input KWCOCO instead of kwiver `CategoryHierarchy`
             import kwcoco
-            paths_to_fix = [self._training_file, self._validation_file]
-            for fpath in paths_to_fix:
-                fpath = ub.Path(fpath)
-                if fpath.exists():
-                    dset = kwcoco.CocoDataset(fpath)
-                    dset.conform()
-                    dset.dump()
+            train_fpath = ub.Path(self._training_file)
+            if train_fpath.exists():
+                train_dset = kwcoco.CocoDataset(train_fpath)
+                train_dset.conform()
+                train_dset.dump()
+                print(f"[MITYoloTrainer] Conforming training dataset")
+
+                kwcoco_cats = train_dset.dataset.get('categories', [])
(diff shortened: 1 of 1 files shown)


ltetrel:
@Erotemic if you have time to review don't hesitate, otherwise I think we are safe to just merge it directly.

Erotemic:
Does it make sense to provide an option for if you want categories from the kwcoco file or kwiver? 

In general I like the idea that a kwcoco file is the one-source-of-truth for categories.

FYI, this will error if not every category has a name, which probably won't happen, but its good to be aware of. 

Should we assert that the train categories have overlap with the validation categories? Or should they be forced to be the same?

Maybe a debug message about what `self._categories` was beforehand, so if someone was relying on "kwiver" categories it becomes easy to see that you had it right previously, but now something is clobbering it?

There's probably a redesign that makes this not even a problem, but I don't have the full context for where `self._categories` is originally coming from and why. 

Overall, if it fixes the problem I think it seems fine.

mattdawkins:
The current paradigm used elsewhere is that if a labels.txt is specified, the detector should train against those categories specified in the labels.txt. If they are not, then it should be trained over all categories in the annotation files.

The purpose of the labels.txt is to over-ride all categories in the annotations file, or merge multiple categories into the same one.

(If you are deviating from that procedure, then it probably shouldn't land as-is)

ltetrel:
Yep I agree with @mattdawkins, we need this framework agnostic label selection from kwiver (actually darknet format). The question is more on making sure that all ML backend create their label mapping file (e.g. for mit-yolo, it is the `train_config.yaml`) based on categories from the dataset (COCO or whatever) to ensure consistency.

ltetrel:
What I am unsure here is if in the original implementation, are kwcoco categories (`training_truth.json`) filter-ed out using the labels.txt ? Seems like not ? @Erotemic @mattdawkins

Erotemic:
@ltetrel See: https://github.com/VIAME/VIAME/pull/268

To make the diff clear, the PR is targeting this branch, so it would need to be merged here, and then this would be merged to VIAME.

PR #268, opened 2026-07-06
Align coco labels with input categories

Erotemic:
@ltetrel

I believe this is a more correct fix. If I'm understanding correctly, the kwcoco files written are temporary, which means they should respect the class order given in labels.txt. This patch ensures the kwcoco files are written with the correct order.

I'm I'm mistaken and the kwcoco files here are not temporary, then this is a bad fix. I already don't like it because it reads the kwcoco files and then writes them out again, which is inefficient. We should probably just be writing them out correctly the first time, but that is a deeper change in code I'm less familiar with. As long as my assumption about the kwcoco files being temporary is true, then this should be robust.

### DIFF
--- a/plugins/pytorch/kwcoco_train_detector.py
+++ b/plugins/pytorch/kwcoco_train_detector.py
@@ -13,6 +13,7 @@ import random
 from PIL import Image
 import scriptconfig as scfg
 from distutils.util import strtobool
+import types
 
 # from kwiver.vital.types import Image
 from kwiver.vital.types import ImageContainer
@@ -22,6 +23,10 @@ from kwiver.vital.types import BoundingBoxD
 
 from kwiver.vital.algo import TrainDetector
 from viame.pytorch.utilities import safe_crop
+
+
+if types.TYPE_CHECKING:
+    import kwcoco
 
 
 class KWCocoTrainDetectorConfig(scfg.DataConfig):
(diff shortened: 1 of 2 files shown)


ltetrel:
I think that should work, can you confirm that you have no label mismatch after training ?
Not sure what you mean by temporary, the written kwcoco files (i.e. `deep_training/training_truth.json` and `deep_training/validation_truth.json`) are used by the training routine from mit-yolo.

ltetrel:
Also, can't this live somewhere deeper (in `KWCocoTrainDetector`) ? So this fix is also applied for other training backends ?

Erotemic:
> Not sure what you mean by temporary

I mean that the kwcoco files written for training are generated via VIAME's source of truth, i.e. after you are done training you could delete them, and they are not used outside of training. 

> Also, can't this live somewhere deeper (in KWCocoTrainDetector) ?

Yes, that makes sense. As a note to myself: the logic in `_align_kwcoco_categories` is probably something that belongs in kwcoco proper. `normalize_category_ids` is almost this, but it lacks the ability to add / remove in addition to reorder.

I moved the `_align_kwcoco_categories` to kwcoco_train_detector.py but the actual usage of it in other kwcoco trainers didn't seem immediately obvious given that other trainers are not handling kwcoco files in the same way MIT-YOLO is.

ltetrel:
Don't understand what happened here. I just merged the PR https://github.com/VIAME/VIAME/pull/267, it automatically deleted the branch, and hence this PR as well.
@Erotemic can you rebase on main instead ? Also have you been able to test if there are still labels missmatch after your changes ?

Erotemic:
Oh, it looks like it was merged directly into main instead of this branch. I thought I had it to point here. Not sure if I messed it up, or it got changed somehow.

I have not tested this end-to-end. I don't have an environment setup to handle that quickly.

We either need to roll main back, or accept the merge and build on top. Having a CI test for this would be good, but I'm not sure how feasible that is or to what degree other training algorithms are tested in CI.

ltetrel:
FYI on our side we made lot of improvement regarding deployment and CI, PR from @Louis-Pagnier-KW here: https://github.com/VIAME/VIAME/pull/251. Although we don't plan to work on CI/CD for other ML backend than LitDet.

PR #272, opened 2026-07-08
Add MINIMA-LoFTR for automatic cross-modality alignment

romleiaj:
# Auto Align via LoFTR

**What:** A Python inference backend that computes the homography aligning two
camera images of the same scene — usually different modalities (EO/RGB ↔ thermal
IR / UV) — using the vendored **MINIMA-LoFTR** deep matcher. It returns a 3×3
homography plus the best ~20 inlier correspondences, or an actionable failure code.

**Why:** EO and IR look nothing alike to classic detectors (SIFT/ORB), so
aligning them meant tedious manual point-picking, especially in the KAMERA system
with its multiple modalities (EO/IR/UV). MINIMA-LoFTR is fine-tuned for
cross-modality matching. The correspondences drop into the existing calibration
flow as ordinary point pairs, and unmatchable scenes fail loud instead of
producing a bad warp. These point pairs can then be directly fed into DIVE
for multicam datasets to link pan and zoom for easy viewing.

**Where:**
- `plugins/pytorch/minima_loftr/` — vendored matcher (inference only)
- `plugins/core/interactive_alignment.py` — matching, RANSAC/refit, quality gate
- `plugins/core/interactive_service.py` — hosts it alongside segmentation +
  stereo in one process; routes `auto_align` / `get_alignment_status`
- GPU memory is freed **on demand** — the matcher is unloaded (model
  dropped, CUDA/MPS cache emptied) just before a competing segmentation/stereo
  model runs. It reloads lazily on the next `auto_align`.

## Verification

9/9 test EO/IR image pairs were successful from a calibration flight - high inlier count with <2px reprojection error. Only took a few seconds on M1 MAC.

### DIFF
--- a/plugins/core/CMakeLists.txt
+++ b/plugins/core/CMakeLists.txt
@@ -455,6 +455,11 @@ if( VIAME_ENABLE_PYTHON )
     interactive_stereo )
 
   kwiver_add_python_module(
+    ${CMAKE_CURRENT_SOURCE_DIR}/interactive_alignment.py
+    core
+    interactive_alignment )
+
+  kwiver_add_python_module(
     ${CMAKE_CURRENT_SOURCE_DIR}/interactive_service.py
     core
     interactive_service )
--- a/plugins/core/interactive_alignment.py
+++ b/plugins/core/interactive_alignment.py
@@ -78,14 +78,17 @@ def find_alignment_weights() -> Optional[str]:
 def _resolve_device(device: Optional[str]) -> str:
     import torch
 
+    if device is not None and device.startswith("cuda") \
+            and not torch.cuda.is_available():
+        # The host service defaults to cuda; degrade to the best available
+        # backend rather than silently landing on CPU.
+        device = "auto"
     if device in (None, "", "auto"):
         if torch.cuda.is_available():
             return "cuda"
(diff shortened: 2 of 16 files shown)


mattdawkins:
Are the models not loaded until the functionality is needed / selected in DIVE, even if there's an existing segmentation or stereo interactive process loaded or job run?

romleiaj:
Right now the LoFTR model is lazy-loaded when the user first hits the "Auto Align" button - it unloads whenever any stereo / seg commands are called. There is still the case where all 3 could be loaded into memory, if the user used stereo + seg and THEN went to auto-align

mattdawkins:
Looks fine to me. Any reason why it's still a draft PR? Has it been tested end to end? Does it have a model that needs to be formulated as a VIAME Add-On?

romleiaj:
It's been tested end-to-end, but the DIVE branch is still floating. And I'd say it should be a VIAME add-on, the model used is here: https://github.com/LSXI7/storage/releases/download/MINIMA/minima_loftr.ckpt (should this go on viame.kitware.com ?)

romleiaj:
Linked with https://github.com/Kitware/dive/pull/1769 for end-to-end functionality

romleiaj:
Closed in favor of #283

PR #277, opened 2026-07-20
coco: normalize keypoint ids and timestamps for kwcoco compatibility

mattdawkins:
Normalizes the COCO/kwcoco output of the core detection/track writers so it round-trips cleanly with kwcoco-based readers.

## Keypoints
- Writer now stamps each keypoint with `keypoint_category_id` (index into the `keypoint_categories` table) in addition to the existing `keypoint_category` name.
- DIVE's kwcoco reader looks keypoints up by `keypoint["keypoint_category_id"]`, so name-only output previously failed to import keypoints into DIVE. Keeping the name means both id-keyed and name-keyed readers work.

## Timestamps
- Image `timestamp` is now written as an ISO-8601 string rather than a raw float of seconds. kwcoco accepts either an ISO-8601 string or a numeric UNIX timestamp; the string form is unambiguous and portable.
- The track reader now parses **both** ISO-8601 and numeric timestamps, so existing files still load.

## Files
- `plugins/core/utilities_coco.py` — `keypoint_category_id` stamping; `seconds_to_iso8601` / `timestamp_to_seconds` helpers
- `plugins/core/write_object_track_set_coco.py` — emit ISO-8601 timestamps
- `plugins/core/read_object_track_set_coco.py` — parse ISO-8601 or numeric timestamps

Field spellings (`timestamp`, `video_id`, `frame_index`, `track_id`, `bbox` xywh, `score`) already matched the kwcoco spec and are unchanged. Verified round-trip on the timestamp helpers and keypoint id stamping.

### DIFF
--- a/plugins/core/read_object_track_set_coco.py
+++ b/plugins/core/read_object_track_set_coco.py
@@ -28,7 +28,7 @@ import json
 from kwiver.vital.algo import ReadObjectTrackSet
 import kwiver.vital.types as vt
 
-from viame.core.utilities_coco import annotation_to_detection
+from viame.core.utilities_coco import annotation_to_detection, timestamp_to_seconds
 
 
 class ReadObjectTrackSetCoco(ReadObjectTrackSet):
@@ -191,8 +191,9 @@ class ReadObjectTrackSetCoco(ReadObjectTrackSet):
                     ann, categories, image_dims=dims,
                     kp_cat_names=kp_cat_names)
 
-                if timestamp is not None:
-                    time_usec = int(timestamp * 1e6)
+                seconds = timestamp_to_seconds(timestamp)
+                if seconds is not None:
+                    time_usec = int(seconds * 1e6)
                 else:
                     time_usec = frame_index
 
--- a/plugins/core/utilities_coco.py
+++ b/plugins/core/utilities_coco.py
@@ -392,21 +392,59 @@ def build_image_list(images, aux_image_labels, aux_image_extensions):
 
 
 def _collect_keypoint_categories(annotations):
-    """Extract unique keypoint category names from annotation dicts.
+    """Build the keypoint_categories table and stamp each keypoint with its id.
 
-    Returns a list of ``{id, name}`` dicts, or *None* if no keypoints
-    are present.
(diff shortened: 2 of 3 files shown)


kelseylmartin:
General Context
There are dozens of groups around NMFS using nearly as many tools to annotate data and various models to generate predictions. Those annotations are on imagery and video, and are variously bboxes, segmentations, point annotations, etc etc. The models are VIAME pipes, Ultralytics family, SAM, and a whole gamut of fully custom architectures from University partners.

We would like to encourage folks to get their input annotations and output predictions into a common format, to empower interoperability. For example, PIFSC has annotations for snapper and grouper species that might be useful for SEFSC models, and the Caribbean has coral annotations that might be useful for PIFSC models. Since all of the data are in sundry formats, converting N formats back-and-forth between N formats requires O(N^2) conversion utilities. Converting N formats into a single common format only requires O(N) conversion utilities. While no format is perfect, it seems as though KWCOCO supports a broad range of the sorts of data and annotations that we need to support, is backward-compatible with COCO, can handle both input and output which is useful for a human+AI-in-the-loop iterative annotation process, and is maintained by our trusted partner Kitware. For this reason, the common format that we are currently encouraging folks to put their data into is KWCOCO.

Several groups are using DIVE to generate annotations and to review model output. That model output may come from VIAME or elsewhere. By encouraging folks to adhere to a common format, we can ease the ability of folks to use DIVE regardless of where their input or output data derive.

Currently DIVE doesn't seem to support KWCOCO out-of-the-box for all of these various use cases. Thus, the transition between our data warehouse and DIVE is cumbersome, and requires an extra conversion step(s). We have not yet begun to attempt to tackle converting from KWCOCO into something that DIVE can use, but have just now begun to scratch the itch resulting from this friction between KWCOCO and DIVE.

Questions
So, we at NMFS need to "support" a KWCOCO and DIVE ecosystem broadly in some sense. Given our context above, do you have any thoughts/recommendations on how we can achieve our goal of KWCOCO input/output and the use of DIVE as a common platform for creating KWCOCO-format annotations and reviewing KWCOCO-format model output? We are interested in your opinions on the various options in terms of valueAdd per unit work.

@brendatran @csbrown-noaa

PR #278, opened 2026-07-21
Consume DIVE camera registration JSON in multi-camera pipelines

romleiaj:
Use the output of the new camera registration process (registration JSONs) in pipelines that require multi-camera alignment. Completely pivots to JSON format for image registration files, and removes support for h5 (and ITK in general) from VIAME.


romleiaj:
Multi-camera pipelines could not use registration produced in DIVE; transforms came from ITK h5 files at paths hardcoded in the pipes.
 
  Adds dive_transform_io, a transform_2d_io implementation reading and writing DIVE camera registration JSON (format v2), wired into auto_detect_transform so a pipe can point at a registration file without declaring its type. Adds 
  warp_detections, a core process that pushes detections through a loaded transform. Ships arctic seal 2-cam and 3-cam detector pipes that run from DIVE against a registration file rather than a fixed path.
 
  The reader consumes the fitted matrices and ignores the observation points, which travel for DIVE's own refitting. Pairs with Kitware/dive#1780, which writes these files from the DIVE side.

mattdawkins:
What in the transformation is specific to dive? isn't it just a .json with a homography in it? I think you need a better name for the dive_transform_io class to something else more descriptive of its purpose (homography_json_io...) if it only loads homographies

mattdawkins:
Going to just rename that on main

PR #282, opened 2026-07-28
Take the stereo calibration from DIVE instead of a global directory

mattdawkins:
### Problem

The measurement and ifremer stereo pipes pointed their calibration at a `global:input_cameras_directory` defaulting to `/home/<user>/Desktop/...`, exposed as a DIVE folder picker.

That picker was a no-op. `$CONFIG{}` expands at parse time with backward references only, while `-s` blocks are appended last, so overriding the global never reached the consuming process — runs silently used the hardcoded path.

### Changes

- Name the consuming process keys directly and declare them in a `# Calibration Keys:` header (needs Kitware/dive#1814)
- Route `ocv_stereo_disparity` and the ifremer `ocv_pair_stereo_*` processes through `read_stereo_rig`, which already handles both files and OpenCV directories, so they accept `.json`/`.yml`/`.npz`/`.mat` — DIVE hands down a single file, but these only ever read a directory of `intrinsics.yml` + `extrinsics.yml`
- Rename their `cameras_directory` config to `calibration_file`, matching every other camera file parameter
- Those formats carry no rectification transforms, so `stereoRectify` now derives them on the first frame, where an image size is finally known
- Image-producing pipes default their output directory to `.`

### Verification

- `kwiver pipe-config` confirms `-s` now reaches `depth_map:computer:ocv_stereo_disparity:calibration_file`, and that the output template bakes to `./depth_map%06d.png`
- `viame_opencv` and `viame_processes_opencv` build and link clean
- DIVE's discovery resolves the expected keys for all six calibration-requiring pipes

### Note

`calibration_params.conf` still has `/home/<user>/...` paths for `video_filename` / `tracks_left|right` — those are example *inputs* DIVE overrides per-run, left as documentation.

### DIFF
--- a/configs/add-ons/ifremer/calibration_params.conf
+++ b/configs/add-ons/ifremer/calibration_params.conf
@@ -6,7 +6,8 @@
 # =============================================================================
 
 config global
-   :output_directory                           /home/<user>/calibration_results/
+   # Trailing slash required: consumers concatenate this directly onto a filename.
+   :output_directory                           ./
    :target_width_value                         9
    :target_height_value                        6
    :square_size_value                          10.
--- a/configs/add-ons/ifremer/measurement_detect_fish_only.pipe
+++ b/configs/add-ons/ifremer/measurement_detect_fish_only.pipe
@@ -2,6 +2,9 @@
 # Title: Stereo detect fish
 #
 # Description: Pipeline for measuring object dimensions in imagery.
+#
+# Requires Calibration: True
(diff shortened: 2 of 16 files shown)


Louis-Pagnier-KW:
@mattdawkins I also added an option to export depth map instead of disparity.
And another option to choose if the exported image is rectified or "un-rectified" to compare depth on original images.
Both parameters are compatible with the `export_as_alpha` setting.

I also confirm that the calibration part you made is working nicely.

## DIVE UI
image

## Rectified depth map visualization
output

## Un-Rectified depth map visualization 
output2

Note:
> (The script for visualization is not part of the PR)

PR #283, opened 2026-08-14
Align cameras via pipeline

romleiaj:
Run the MINIMA-LoFTR model on a subset of imagery, rather than a single pair. Run via pipefile (2cam or 3cam pipelines) rather than a service.


mattdawkins:
Mostly seems okay - though the common 2 and 3 input camera pipelines should probably not be in the add-ons and instead just be in the root configs folder. They are also duplicative with the stereo inputs so should be merged to not have duplicate pipelines (the common_input_stereo.pipe could probably just be removed in favor of the 2-cam input pipe, if they are truly the same, and includes updated).

I think some of your pipelines operate off 16-bit imagery, aren't the common 2 and 3 cam inputs converting to 8-bit naturally with the force byte flag on to true, is that intended?

romleiaj:
Got it, moved the input pipes to the root configs folder.

- common_stereo_input.pipe → configs/pipelines/common_2cam_input.pipe, content unchanged, with its one includer (common_stereo_input_with_downsampler.pipe) updated.
- common_3cam_input.pipe moved to configs/pipelines/
- The add-on now holds only the two utility_align_cameras_*.pipe.

They weren't quite identical though, the 2cam pipe carried two extra settings, and neither is compatible with stereo:

- disable_image_load true — common_stereo_input_with_downsampler.pipe connects input1.image/input2.image into its downsampler, so this would hand it null image containers, affecting the 18 pipelines downstream of that wrapper.
- no_path_in_name false — feeds track_writer:image_file_name, the image-name column of the output CSV (e.g. measurement_calibrate_cameras_default.pipe:89), switching basenames to full paths.

So common_2cam_input.pipe keeps the stereo defaults verbatim, and both settings moved into the align utility pipes as post-include config input1 / config input2 overrides.

I left common_stereo_input_with_downsampler.pipe and _with_tracks.pipe under their current names, renaming would touch 18 files across seven add-ons for no functional gain, can address now if you'd prefer.

On `force_byte`: it does crush the image but the pipes never read that image. They connect only the file_name ports, and the process re-opens each file itself at native depth and applies a percentile stretch. Added disable_image_load true so those frames aren't decoded in the video loader just to be ignored.

PR #288, opened 2026-09-08
feat: Add measure pipeline based on head/tail + disparity map

Louis-Pagnier-KW:
Add a new stereo pipeline to measure fish based on head/tail keypoints + disparity map.
Currently the existing measure process is based only on triangulation from keypoints, but if these keypoints are mislocated a bit or a detection in partial, it will return very big values.

So instead, this PR algorithm search for depth in the disparity map, and then compute 3d length based on this.
To give more robustness, multiple points are sampled along the head/tail segment to detect outliers in the disparity map and avoid to pick a background pixel.

Also, the process match fish and associate track ids, so in DIVE when a fish is selected, the corresponding is highlighted in the other view.


mattdawkins:
Pipeline filename itself should be more descriptive to differentiate it from other pipelines that also do measurement based on head/tails

Louis-Pagnier-KW:
I renamed it to `measurement_from_head_tail_disparity.pipe`, is it better ?

mattdawkins:
maybe measurement_from_annotations_unrect_disparity.pipe?

Technically all of our pipelines right now use head/tail, so that's an empty designation.

Is the purpose of this pipeline to do measurement on annotations in the viewer that the user has already made and generated and is shown in the UI? If so, that's what the other "from_annotations" pipelines do. Is it expected that the actual loaded imagery is an unrectified disparity map, or just that you have one in the folder?

mattdawkins:
Bryon and I were also thinking about renaming measurement_* to stereo_* then we could have the measure in there still to differentiate calibration generation and test pipelines from measurement if no objection

Louis-Pagnier-KW:
The pipeline is intended to be used both, on annotated frames but also automatically generated head/tail keypoints using SAM2 (or any other head/tail extractor) as it is robust to some errors in these keypoints placement.

The expected input is just a normal stereo image pair, the pipeline compute itself rectification and disparity.
So may be "from_annotations" is not the best prefix.
Something like `measurement_unrect_disparity.pipe` ?

But yes of course, it will better fit it's name after the stereo/measurement refactor

mattdawkins:
Just went ahead and did the stereo rename so this needs a rebase now.

>>  but also automatically generated head/tail keypoints 

But that's not in the pipeline right, (no detector) - you're still just relying on whatever annotations are on the current dataset?

If so the newer specification for those would be: stereo_measure_current_annots_[algorithm].pipe so something like stereo_measure_current_annots_unrect_disp.pipe if going by that convention. We already have an existing process for this, however, so I wouldn't recommend that name now, nor creation of a new process in this branch (it's used in the foundation stereo model add-on, which is also mapping based on unrectified disparity maps). So going further if you want to keep this branch as-is I would call it stereo_measure_current_annots_sgbm.pipe to differentiate (or whatever you're using for the disparity algorithm)

I've refactored this a little bit on PR https://github.com/VIAME/VIAME/pull/294 to use the existing process fast fdn uses

If you agree with those changes let me know and we can abandon this one

PR #294, opened 2026-09-11
Add robust disparity segment fitting to shared stereo measurement

mattdawkins:
Adds robust head-tail disparity fitting to the existing `compute_measurements` process. This is an alternative to #288 for comparison; #288 is unchanged and its original head is also saved as `backup/pr-288-before-shared-measurement-20260910`.

The shared process samples disparity along the annotated segment, rejects invalid samples and outliers, and fits disparity versus image position before triangulating endpoints. This avoids fitting XYZ coordinates against image interpolation fractions, which biases lengths under perspective. It reuses existing stereo track association, rectification, output writers, and track-length aggregation rather than introducing another Python process. Rejected frames do not replace prior valid track measurements with zero.

- Adds `stereo_measure_current_annots_sgbm.pipe`, using SGBM/WLS and robust refinement.
- Enables the new refinement only in the SGBM pipeline. Foundation Stereo pipeline files and default sampling behavior remain unchanged.
- Validates sample counts, outlier limits, and residual thresholds. Fits require at least three inliers, a strict majority, and support spanning half the segment.
- Uses correct pixel-stride addressing in the new segment mode. Retains legacy endpoint sampling outside that mode to avoid changing existing Foundation Stereo results; a broader sampler correction is outside this PR.
- Documents configuration, calibration units, and the straight-segment assumption.

Validation: compiled changed native sources and linked the core library and process plugin using the existing build configuration in an isolated output directory. All 58 measurement utility tests pass, including new perspective, outlier/missing-data, configuration, disparity-format, reversed-endpoint, rectification/triangulation, and legacy-sampling compatibility coverage. The new SGBM pipeline passes `viame pipeline check`. No full clean build, model-backed run, or before/after dataset comparison was performed.

### DIFF
--- a/configs/add-ons/fdn-stereo/stereo_measure_current_annots_fdn_stereo_s.pipe
+++ b/configs/add-ons/fdn-stereo/stereo_measure_current_annots_fdn_stereo_s.pipe
@@ -49,6 +49,14 @@ process measurer
   :refine_keypoints_with_disparity             true
   :refine_keypoints_disparity_window           7
 
+  # Optional robust profile fit, shared with the SGBM measurement pipeline.
+  # Works with foundation_stereo and fast_foundation_stereo disparity backends.
+  # Invalid samples and outliers must leave >=3 samples and a strict majority.
+  :refine_disparity_segment                    false   # DIVE_PARAM["Robust Segment Fit", bool]
+  :disparity_segment_samples                   11      # DIVE_PARAM["Segment Samples", range_int, 3, 101]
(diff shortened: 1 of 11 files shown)


Louis-Pagnier-KW:
image

It looks like the algorithm itself works, but at the end it seems to aggregate all tracks on the first frame.

Louis-Pagnier-KW:
I found the bug, I will push the fix tomorrow

Louis-Pagnier-KW:
I had to change a few things in the pipeline so the algorithm is similar to the python version.  
Now it is working as intended.

